PR #33, opened 2022-10-11
Correct some typos in README.md

Peter0x44:
I noticed some typos in README.md so I went ahead and fixed them
I checked over the rest of the file with vim `:spell` option and noticed nothing else wrong.


skeeto:
Doh! Thanks! I should configure a "set spell" autocmd for Markdown so that 
I don't forget to turn it on.

PR #34, opened 2022-10-19
Update make 4.2 -> 4.3

HuangShumin:
request to update make version


skeeto:
Upgrading requires updating the hash as well. I just pushed a branch that 
does this, make-4.3 (currently 2c1946f). Cross compilation is broken in 
GNU Make 4.2, so I had a hack to fix it (-Iglob). That seems to be fixed, 
so the hack should be removed, too.

Unfortunately GNU Make 4.3 is incompatible with Mingw-w64, so this still 
does not work. Otherwise I would have done it already. This version makes 
assumptions about fcntl() and dirent that don't hold in Mingw-w64, and so 
the build fails. (See my notes in the commit.) It would probably take 
substantial patching to fix.

It's quite possible that w64devkit will forever have GNU Make 4.2. "The 
devil you know" and all that. It's a feature-complete mature program, and 
there is practically no innovation happening in this space. The only 
significant future improvements would be better Windows support (e.g. 
learning how to deal with PowerShell for SHELL, and its limitations). More 
likely I'd replace it altogether with a different make that has reasonable 
GNU Make compatibility, enough to work with existing, real world GNU 
Makefiles (in Vim, Putty, zlib, etc.).

skeeto:
GNU Make 4.4 recently came out, and this version seems to work well on Windows, so I've merged it as 1000d8e. (The new `--shuffle` feature *is* intriguing.)

PR #48, opened 2023-02-05
Fix 404 Win32 link

danielmartin:
The file is available via the Internet Archive, luckily.


skeeto:
Thank goodness for the Internet Archive. Thanks, Daniel!

Issue #54, opened 2023-04-06
Issues with `ld` crashing


luc0x61:
I've had issues with `ld` **crashing** on invocation, which I resolved using a linker from another MinGW binary distribution. Unfortunately, I can't share the setup or the files involved, then I understand how *vague* is all this, even though I started a tentative explanatory question on [StackOverflow](https://stackoverflow.com/questions/75888792/is-there-a-very-verbose-output-for-ld-linker-linker-crashing-silently).

I replicated the problem in a couple of environments. It appeared/disappeared even after changing a few instructions in the sources. Something suggests an issue with the kind of initialization of the memory, as at a certain point I had also different results in different command processors (cmd.exe, TCCle).

Peter0x44:
Can you at least share the version of binutils the other working binary distribution was using?

skeeto:
Looks like I hadn't noticed a Binutils release earlier this year, so I've 
updated that to Binutils 2.40 in 7f6e0ba. The release notes indicate it 
contains "numerous bug fixes" which perhaps includes your issue. If you're 
able to build w64devkit yourself — requires only Docker or Podman — give 
that a shot. Since you're using the i686 variant, use my script:

    ./multibuild.sh -q4

This works with Docker Desktop on Windows, and you can run the build shell 
script using w64devkit itself. That will produce a w64devkit-i686.zip for 
you to try. Alternately you could provide a minimal example so that I 
could test it. I was unable to produce the crash using --image-base.

It's now possible to build Binutils using w64devkit (see #50), so another 
approach is to compile a Binutils debug build, then point w64devkit's GDB 
at it to examine the issue in a live processes with debug symbols. If I 
could reproduce the bug I'd try this myself.

I don't anything particularly special building Bintils aside from:

* --with-static-standard-libraries
* PE timestamps disabled by default

These seem unlikely to affect your issue, especially since you're using 
"-static" explicitly, so I suspect the other ld also has the bug but just 
so happens to be built in such a way it doesn't crash.

Peter0x44:
Hopefully newer binutils versions fixed whatever this problem was.
If not, feel free to open another issue.

PR #59, opened 2023-05-01
Fix a regression in Binutils 2.40

GalaxySnail:
In Binutils 2.40, there is a regression that ld fails with MSVC lib files. Add a patch to fix it.

[1] https://github.com/msys2/MINGW-packages/issues/15469
[2] https://sourceware.org/bugzilla/show_bug.cgi?id=30079

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -152,7 +152,9 @@ RUN make -j$(nproc) \
 # Cross-compile GCC
 
 WORKDIR /binutils
-RUN /binutils-$BINUTILS_VERSION/configure \
+COPY src/binutils-*.patch $PREFIX/src/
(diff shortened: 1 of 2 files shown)


skeeto:
Thanks, GalaxySnail! The bugzilla report included a good reproduction 
sample, and the patch does indeed fix it. I shifted the patch up the 
Dockerfile to merge into my earlier sed-based "patch." I expect this is a 
short term fix and will be deleted the next time I update Binutils.

I wonder if this is connected in any way to #54.

GalaxySnail:
> I wonder if this is connected in any way to #54.

This regression first appeared in binutils 2.40, while binutils 2.39 worked well. So I think #54 may be a separate problem.

PR #66, opened 2023-05-16
Add GnuPG

Semphriss:
I did this for my personal use, but I thought I'd share, in case it might be useful.

I noticed GnuPG supports some extra libraries that are optional (such as ntbTLS); I didn't bother compiling them in. I'm not sure what's the exact impact of leaving them out though, so it might be good to verify.

If there's interest in merging this, and some changes are needed before merging, I'm open to update my PR as necessary. I also leave it open for modification by maintainers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,18 +9,24 @@ ARG EXPAT_VERSION=2.5.0
 ARG GCC_VERSION=13.1.0
 ARG GDB_VERSION=13.1
 ARG GMP_VERSION=6.2.1
+ARG GPG_VERSION=2.4.1
+ARG LIBASSUAN_VERSION=2.5.5
+ARG LIBGCRYPT_VERSION=1.10.2
+ARG LIBGPGERROR_VERSION=1.47
 ARG LIBICONV_VERSION=1.17
+ARG LIBKSBA_VERSION=1.6.3
 ARG MAKE_VERSION=4.4
 ARG MINGW_VERSION=11.0.0
 ARG MPC_VERSION=1.2.1
 ARG MPFR_VERSION=4.1.0
 ARG NASM_VERSION=2.15.05
+ARG NPTH_VERSION=1.6
 ARG PDCURSES_VERSION=3.9
 ARG CPPCHECK_VERSION=2.10
 ARG VIM_VERSION=9.0
 
 RUN apt-get update && apt-get install --yes --no-install-recommends \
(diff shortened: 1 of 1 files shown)


skeeto:
Thanks for sharing! I'm not interested in including GnuPG, but I'm excited 
to see that you've modified the build for your own needs. One of my goals 
was to make that kind of modification possible and relatively easy.

I'm not necessarily opposed to including gpgv.exe, the standalone 
signature verification tool. Unlike the rest of GnuPG, it's well-behaved, 
self-contained, and doesn't interact with gpg-agent. It could validate the 
next w64devkit release to be installed, which have always been signed. It 
would also fit with the idea of verifying third-party libraries to be used 
in builds. However, it would still require practically all the same steps 
you've taken to build it, including dependencies, and it's simply not 
important enough to warrant that.

Personally, I view OpenPGP encryption as a technological dead end — no 
forward secrecy, etc. — and so it's not a capability worth having. As a 
signature format it still has value because, unlike encryption, it's 
(mostly!) not compromised by backwards compatibility, and no other format 
(minisign, etc.) has gained the mindshare to replace it in the general 
case.

A small signing tool would be nice to have, say, for signing software 
releases built with w64devkit. Generating OpenPGP signatures is super 
easy, and I've considered porting a subset of my passphrase2pgp tool to C, 
which would fit well in w64devkit and sidestep keychain management. Though 
in this context, signtool-compatible signatures would be more useful and 
appropriate, e.g. osslsigncode. A few times I've considered writing one 
myself.

As for GnuPG itself, I dislike its architecture. gpg-agent is probably at 
its worst on Windows, continuously holding mandatory locks on the files 
and directories it touches. If the OpenPGP ecosystem wasn't effectively 
obsolete, I'd love to see a clean, slim alternative more in the style of 
gpgv. OpenPGP is a complex format with lots of options, and the hard part 
of replacing GnuPG is supporting a substantial portion of real-world 
inputs, which requires implementing and parsing all those options — a lot 
of work.

That being said, I *do* distribute a GnuPG build for Windows separately:
https://github.com/skeeto/gnupg-windows-build

I don't use it much, but I do have it "installed" on my PATH. I also view 
supplying a tool to validate w64devkit signatures as complementary. It's 
compiled with Debian's Mingw-w64 distribution rather than w64devkit's 
bootstrap compiler, as is the case with your patch, but you probably 
couldn't really tell the difference in the end.

Semphriss:
Ah, I didn't see the GnuPG repo! I'll give it a look.

I'll close this PR, my implementation has many issues.

PR #77, opened 2023-07-11
Add Git

Semphriss:
I hope my opening of many PRs lately don't come off as burdensome, please do tell me if my contributions are excessive. I'm opening this one because I noticed the second paragraph in [Notes](https://github.com/skeeto/w64devkit#notes) mentioning an interest in Git, and I've succeeded in coming up with a somewhat simple solution (Git only depends on zlib and curl).

---------------

This adds a relatively fully functional Git to w64devkit. I've taken quite a few decisions during the production of this patch, and I'm willing to update any of them as preferred.

<details>
<summary>Decisions</summary>
<ul>
<li>The curl dependency is technically optional, but without it, Git doesn't support <code>http[s]://</code> origins, which I assumed was an important feature.</li>
<li>curl is dynamically linked; I'm not sure why Git doesn't want to pick up the static builds.</li>
<li>No SSH client is built, because Windows comes with one already. Wine does not, so ssh origins do not work under Wine.</li>
<li>curl supports a variety of SSL libraries; I compiled it with the one that said "Windows native SSL/TLS", which I figured would be the most lightweight and the simplest option on Windows. I did not test Git with Windows XP.</li>
<li>No localization, no GUI and no documentation, to keep the build as simple and small as possible.</li>
<li>Git is configured to look for <code>less.exe</code> as the pager and <code>vi.bat</code> as the editor. If none are specified at build time, Git would look for "less", find no match (less is "less<b>.exe</b>"), and fallback on simply printing everything, and for the editor, look for "vi", find no match for similar reasons and fail.</li>
<li>Git's <code>make</code> and <code>make install</code> are done in one command, because both require having the same arguments, so doing both separately would duplicate all the arguments.</li>
<li>Git generate one executable for each command, but the commands are 3 megabytes each, and <code>zip</code> doesn't seem to support symlinks very well, even with <code>--symlinks</code>, so I used the <code>alias.c</code> program to emulate those symlinks.</li>
</ul>
</details>


Kreijstal:
why 32bit

avih:
Hmm.. is this a pure mingw git? is there an existing pre-built binary/package of pure mingw git? I'm not familiar with such package.

As far as I know the official git-for-windows doesn't have a pure mingw build. Even the minigit busybox build still has msys2 binaries/dll for some https stuff IIRC.

Regardless, even if git is useful in the context of w64dk (and it definitely is), I don't know if it belongs in w64dk. I'd think it would be more appropriate to distribute it individually, independently of w64dk. Kinda like git-for-windows, but a pure mingw package if possible.

PR #99, opened 2023-12-10
Delete UNLICENSE

fiiiiiiiiiuzzz:


install.ps1
FinnT730, line 1: bit weird that your powershell script is a binary....

PR #104, opened 2024-01-17
Codespace miniature space system qwj4p7jpq59c95q

fiiiiiiiiiuzzz:


FinnT730:
...
Can someone just kick / block this guy from this repo?

The binary Powershell scripts are Sus, and weird PR names too, which have nothing todo with the repo...

At this point, its just spamming the repo for no real cause

FinnT730:
Thanks. Yeah, I hate the idea of blocking someone from a repo, but this is not the first time they are trying to get rid of the license and including a shady powershell script, which would likely be malware. Also, to be fair, for a sample java project, I can see why the PR's can be silly etc. This is something bigger :P

PR #120, opened 2024-03-31
Upgrade to expat-2.6.2

ducalex:
The old release is no longer hosted upstream so the build fails.


skeeto:
Thanks, Alex!

Perhaps I should figure out some way to get these from GitHub, because 
these fossies.org links have been unreliable. That's the second time this 
month it's happened. The problem with the official download URL has been 
that the libexpat devs use weird tag names that don't match the actual 
version number, e.g. R_2_6_2 instead of 2.6.2 or v2.6.2.

PR #126, opened 2024-05-01
Upgrade to Binutils 2.42

Semphriss:
Binutils 2.42 was released in January.

I also checked the rest of the software, and everything should be up-to-date. I saw that GCC has a 14.x branch, but apparently no releases were made so far.


skeeto:
Perfect, thanks, @Semphriss! Took it for a quick test drive, and 2.42 
seems to work just fine.

PR #153, opened 2024-08-02
Add compression options to 7zip

Semphriss:
Using some compression options allow shaving a few extra MB's off the archive.

It also removes the timestamps from the archive, so it might make it easier to make bit-for-bit reproducible builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -497,6 +497,6 @@ RUN printf "id ICON \"$PREFIX/src/w64devkit.ico\"" >w64devkit.rc \
  && cat /mingw-w64-v$MINGW_VERSION/mingw-w64-libraries/winpthreads/COPYING \
         >>$PREFIX/COPYING.MinGW-w64-runtime.txt \
  && echo $VERSION >$PREFIX/VERSION.txt \
- && 7z a -mx=9 $PREFIX.7z $PREFIX
+ && 7z a -t7z -mx=9 -mfb=273 -ms -md=31 -myx=9 -mtm=- -mmt -mmtf -md=1536m -mmf=bt3 -mmc=10000 -mpb=0 -mlc=0 $PREFIX.7z $PREFIX
 ENV PREFIX=${PREFIX}
 CMD cat /7z/7z.sfx $PREFIX.7z


skeeto:
Thanks! I definitely want to drop timestamps like I had been doing with 
the previous zip distribution. I'd like to understand the other parameters 
better rather than cargo cult them from a copy-paste. A few appear to be 
redundant (-md appears twice, and others are already the default), and of 
course it's critical that the 32-bit distribution is decompressible with a 
32-bit decompressor using a reasonable amount of memory. A 1.5G dictionary 
would be straining. (Note: w64dk operates well and smoothly on 20 year old 
computers with relatively low amounts of RAM!)

My local Debian man page is paper thin, but I found this:
https://7-zip.opensource.jp/chm/cmdline/switches/

I'd like to figure out which of these switches gives so much more mileage 
beyond -mx=9, and focus on those. For some parameters it's unclear which 
direction, smaller or larger, results in smaller files (e.g. lc, pb).

Semphriss:
According to my `man 7z` down to the `-m` options, there are exhaustive explanations in `/usr/share/doc/p7zip/DOC/MANUAL/cmdline/switches/method.htm`, which is where I got this from. I essentially just set every option to the value that gives the best compression ratio. I didn't try every combination though, it's possible I made suboptimal choices.

skeeto:
After some experimentation, it seems the "fast bytes" (-mfb=273) option is 
responsible for nearly all the gains, and so 29696b88 is what I'm looking 
to merge. Enabling multithreading (-mmt) makes no difference at all in my 
tests.

There is little documentation for what "fast bytes" means or what it 
affects, but skimming through the 7-zip source… maybe lookback distance or 
match length? It seems it also doesn't increase decompression resources, 
just compression time, so no worries about 32-bit hosts or other smaller 
computers. If you have more insight on this parameter, let me know!

On my system, fb=273 added an extra 40 seconds to the build time, a ~60% 
increase to that step. Considering that this is at the very end and so 
it's never cached while iterating on changes, for a mere 700K (~2%) extra 
compression I'm not sure it's worth it unless I can turn it off most of 
the time.

Semphriss:
I found it surprisingly difficult to find information about the signification of the options online, but I did find two sources, both of which are forum comments on obscure posts, that say the "Fast bytes" are actually just the word size of the dictionary:

- https://stackoverflow.com/questions/39914398/7zip-fastest-lzma2-compression
- https://sourceforge.net/p/sevenzip/discussion/45797/thread/00b4d25b/

You may need to Ctrl+F "fast bytes" to reach the relevant part, they're not obvious. I couldn't find anything more official or credible than that, other than flat reposts of the 7zip manual which doesn't say anything other than the vaguely expected impact.

As for the value, 273 is the maximum, but it might not be really necessary to make it that high. Some sources say 64 is enough. I have a fork with different software, that is >500Mb that I can get down to 47mb without options and 43mb with, and with just `-mfb=???`, I got this:
- 273: Took 2m50s, resulting size is 45909675 bytes.
- 64: Took 1m44s, resulting size is 46623364 bytes.
- 32: Took 1m26s, resulting size is 47762529 bytes.

At this point, I'm not sure if the options are worth it at all.

N-R-K:
[According to Igor](https://sourceforge.net/p/sevenzip/discussion/45797/thread/bebc01be/?limit=25#544f/0d93/3151):

> Match finder checks only this number of bytes.

I have no clue what this actually means since I don't know the specifics of 7zip's match finder. In general, lz77 match finders are not exhaustive, they are heuristic based and have various limits in place for both performance reasons and also to avoid quadratic runtimes. The "fast byte"/"word size" is probably tuning one of those limits.

I also disagree with setting a 1.5G dictionary size. It's redundant since w64dk doesn't even reach that size (it's about 400M) but might be actively harmful since a space constrain system may refuse to decompress such file with a large memory requirement.

skeeto:
I decided to keep just the timestamp change. Since you spotted it, I 
wanted you to have credit as author of the commit, so I did a quick 
modification in place.

Issue #69, opened 2023-06-10
Final Zip File Cannot unzip Error


cuongnguyengit:
![image](https://github.com/skeeto/w64devkit/assets/45245565/4293025f-29b2-4adb-90bc-0f3e669bbef3)

anyone help me, plz?

skeeto:
Can you share the exact steps you used to build this? That .zip is about 
twice as large as it should be. With "hd w64devkit.zip | head" the first 
two bytes should be "PK".

cuongnguyengit:
I set up follow your direction
git clone https://github.com/skeeto/w64devkit.git
cd w64devkit
docker build -t w64devkit .
docker run --rm w64devkit >w64devkit.zip

skeeto:
I thought more about the 2x detail and realized what's happening: You're 
using PowerShell, right? "No way could it screw up something as basic as 
file redirection," I thought, yet I tried it myself and can exactly 
reproduce your issue. This is typical PowerShell brain damage where 
"redirect to a file" actually means "redirect through PowerShell" which 
then "helpfully" translates your data into UTF-16, a format nobody wants 
nor uses for text files. That, of course, increases the size by 2x — a 
familiar outcome. This is not my first rodeo.

The solution: Use either cmd.exe or w64devkit itself. PowerShell is an 
embarrassment of a shell, a kind of cruel joke that keeps on going. One 
reason w64devkit exists is that, even still in 2023, Microsoft with their 
trillions of dollars is unable to produce and supply a decent system shell 
that can compete with hobbyists.

cuongnguyengit:
it was really worked in cmd. Thanks for your support.

garoto:
Timelines must have forked around the late 70s or early 80s and we're now living its worst version possible, where MS in-house garbage tech is seem as "standard" and newer generations can't possibly know any better, since the shilling is _too stronk_.

You should write about your thoughts regarding powershell in your website. Would love to see if my non-computer-programmer concepts can/would somewhat align with the ones from a seasoned C programmer at all.

PR #160, opened 2024-08-18
One-liner powershell script to setup env

dezashibi:
- download the latest version
- install it silently
- add the `bin` folder to user's path if it's not already been added

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ PowerShell because it lacks file redirection.
 To setup and config the environment automatically you can run the following script on Windows Powershell:
 
 ```powershell
-powershell -c "irm https://raw.githubusercontent.com/dezashibi-c/w64devkit/master/install.ps1 | iex"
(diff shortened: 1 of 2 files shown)


skeeto:
Thank you for taking the time to put this together and share it, @dezashibi. I tried it out and it worked exactly as advertised, even with the basic PowerShell that comes with Windows. However, I'm not interested in accepting this pull request.

Modifying the user's PATH is too installer-like and more invasive than my goal. I want a no-admin-needed, leave-no-trace portable software distribution where "uninstall" is merely deleting the w64devkit directory, because no system changes were made. It also supports multiple, parallel installations which don't interfere with one another, which isn't the case with a global install location and PATH changes. Besides that, while *only* adding it to the PATH is a valid way to use w64dk ("just give me a compiler, please"), using it that way misses out on core features like the shell, using Autoconf builds, and "virtual" home directories.

These could be addressed with changes to the script, but that brings up my next reason: I dislike PowerShell and want to have nothing to do with it. There's even a note to the README not to use PowerShell when building w64devkit because it messes up the output in a surprising, non-obvious way (#69).

PR #182, opened 2024-10-30
Update Expat to 2.6.3

Semphriss:
Updated given the few security issues in Expat 2.6.2. I did not test it because I currently don't have my Windows PC nearby, but I'll probably use it in the next few days.


skeeto:
Thanks, @Semphriss! In w64dk, libexpat only parses trusted inputs, so none 
of its security issues are ever relevant. Though the old version no longer 
being hosted *is* relevant, so the upgrade is still required.

PR #183, opened 2024-10-30
Update 7-zip to 24.08

Semphriss:
Similarly to PR 182, I haven't yet had the opportunity to test the changes, but I thought it would be preferable to open to open the PR early. I'll test this whenever I'll have an opportunity.


Semphriss:
I forgot to commit the `7z.mak` changes, I've force-pushed an update. There seems to be just one more object file needed.

skeeto:
I appreciate these kinds of PRs, even if I don't jump on them. As job is especially narrow, the only reasons to update the SFX program is for performance or if it's unpacking improperly due to a bug, and so it's possible I may never update it. But I like that the option is here.

PR #194, opened 2025-02-08
Upgrade to GDB 16.2

Peter0x44:
The patch for main needed some adjustments, but otherwise everything worked in my testing.
Not sure if I changed it correctly but at least it didn't seem to break anything...
I compiled a WinMain entry point program and it was fine.


skeeto:
Thanks, @Peter0x44! It was particularly helpful fixing up the entry point 
patch for me. It's irritating that they change the interface from release 
to release, requiring this maintenance on most upgrades. Perhaps I should 
try to upstream the patch. After a quick test, GDB 16.2 seems to work fine 
including the updated entry point patch.

I fixed up your patch changes to restore the original GNU style mixed tabs 
and spaces. I wrote it that way to match Binutils ("when in Rome…"), not 
because I like it.

(Also, the fossies.org nameserver is down today, so I updated Expat while 
switching the download over to SourceForge so that I could build with your 
new changes.)

PR #196, opened 2025-02-18
gcc: Patch misaligned AVX spills (GCC bug 54412)

skeeto:
I believe this patch fixes [GCC bug 54412](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412), but it's a stab in the dark. I'd like more eyeballs on it, and help testing, especially because it mysteriously works better than I anticipated. That is, GCC still emits aligned moves when it's valid to do so, and it seems this only affects the cases that were broken (i.e. spills). I had expected some valid aligned moves would become unaligned, at probably some negligible (and acceptable) performance cost.

In the worst case I expect that this patch is incomplete, and there are still some misalignment crashes. I do not expect this breaks any code generation. The patch is unsuitable to go upstream as-is because it would affect all i386 targets, but perhaps it could be adapted into a general, upstreamable patch.

The GCC bug report contains a couple of sample programs that no longer crash after my patch. I've also tested llama.cpp (via `contrib/llama.mak`) with `-O2 -march=x86-64-v3`, which GCC would miscompile without `-Wa,-muse-unaligned-vector-move`. With this fix, it works fine without the special assembler flag.

If you'd like to help, for any programs you know that break with `-mavx` / `-march=x86-64-v3` / `-march=native` / etc., try building with this fix and see if it no longer crashes. I want to know if this patch (1) newly breaks anything, and (2) misses existing crashes related to AVX alignment. In case of (2), if I can't identify another keyhole patch I'll probably just restore the note in the README and be happy that at least this situation is less bad.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,10 +183,6 @@ ported to Mingw-w64][san] ([also][san2]), but Undefined Behavior Sanitizer
 `-fsanitize-trap`, GDB will [break precisely][break] on undefined
 behavior, and it does not require linking with libsanitizer.
 
-[GCC does not fully support AVX on Windows][avx] and may use aligned moves
-on unaligned addresses. When targeting AVX, consider disabling all aligned
-moves in the assembler: `-Wa,-muse-unaligned-vector-move`.
-
 ## Licenses
 
 When distributing binaries built using w64devkit, your .exe will include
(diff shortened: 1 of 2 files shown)


Peter0x44:
I think it would be best to leave a comment highlighting this patch on the bug report itself to get eyeballs on it.

skeeto:
Good thinking, @Peter0x44. I allowed my [shyness](https://www.schneier.com/blog/archives/2013/04/nice_security_m.html) to make me flinch from using a resource. It's a bad habit. I've [shared my patch](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54412#c45).

Though while double-checking, I managed to contrive cases producing unaligned loads unnecessarily, which is in line with my expectations. That makes me more confident it does what I think. I also learned that Debian has a similar patch (also shared in the bug threat), and their cross toolchain produces the same results as my patch.

PR #205, opened 2025-03-27
Upgrade to expat 2.7.0

Peter0x44:


skeeto:
Thanks, another easy merge!

Yeah, there's no true reason to upgrade unless GDB requires it, i.e. uses 
a new feature. Expat security is irrelevant because it only parses trusted 
inputs. However, the Expat maintainers' naive security mindset has them 
deliberately trashing the old download URLs, necessitating a response. Not 
all projects are under the same constraints! GitHub-generated tarballs are 
more stable, though that also requires navigating Expat's unconventional 
tagging system.

About a year ago I began exploring vendoring a subset of Expat (6f1f2b9b), 
which sidesteps this whole problem, eliminates a dependency, and builds in 
a fraction of the time. Alternatively I also considered just writing GDB's 
needed Expat subset from scratch, which I expect would take little code.

Issue #86, opened 2023-09-01
ld error: multiple definition of `std::type_info::operator==(std::type_info const&) const`


xparq:
**w64devkit 1.20.0**, both 32/64 bit
  _(Tried to check with gcc12 in an older `w64devkit` which I also have here, but the project that triggers it depends on `<format>`.)_

* Does not happen with a stock Debian stable on WSL, using GCC 13.2 (installed from `testing`): `g++ (Debian 13.2.0-2) 13.2.0`
* Same project builds clean with MSVC.
* It's a single compilation unit.
* I've been using [doctest](https://github.com/doctest/doctest), but it didn't cause any issues earlier, and I don't think I've changed anything related.
* I noticed this error a little while after adding `std::regex` stuff to the code. Not sure yet if that's related, I might test it out, if this doesn't seem familiar from elsewhere/upstream. _(I've seen multiple vaguely similar reports for GCC etc., but the best matches were unanswered SO questions. [Answered [one](https://stackoverflow.com/questions/76297301/multiple-definition-error-when-statically-linking) in the meantime.])_
* The project can afford `-fno-rtti`, so that's a viable workaround for me.

~~~~
g++ -Wall --std=c++23 test.cpp
.../ld.exe: .../w64devkit/gcc13/64/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/libstdc++.a(tinfo.o):tinfo.cc:
(.text$_ZNKSt9type_infoeqERKS_+0x0): multiple definition of `std::type_info::operator==(std::type_info const&) const';
...\Temp\ccOVUCq4.o:test.cpp:(.text$_ZNKSt9type_infoeqERKS_[_ZNKSt9type_infoeqERKS_]+0x0): first defined here
collect2.exe: error: ld returned 1 exit status
~~~~
<sup>&nbsp;&nbsp;_(Edited for readability.)_</sup>

xparq:
Yup, confirmed; this single line triggers it:

    std::regex regx("");

Probably the regex compiler. (Just `std::regex regx();` is not enough.)

skeeto:
This appears to be a bug in libstdc++, perhaps related to static linking, 
which would explain why it's mostly gone unnoticed. (w64devkit's static 
linking libstdc++ is unusual.) I can reproduce it from the cppreference 
std::type_info::operator== example, even when reduced to:

#include <typeinfo>
int main() { return typeid(0) == typeid(0); }

[...]

PR #221, opened 2025-05-08
Delete unneeded patch for #86

Peter0x44:
The issue in GCC (PR110572) has been fixed upstream, so this patch is no longer needed.


skeeto:
Thanks! I confirmed it with a quick build. I was thinking about this patch 
when updating to GCC 15, and I figured something would break if it was no 
longer needed.

PR #250, opened 2025-07-27
Upgrade to binutils 2.45

Peter0x44:
I built w64devkit and tried it out on the following things:
Building+Stepping through a C program
Building+Stepping through a C++ program
Doing a profiling run using `-pg -no-pie` and then using gprof

and they all worked without issue.


skeeto:
Thanks! Looks good to me, too. I amended the commit to include the release 
announcement so that it's on hand later when I write my own release notes.

PR #272, opened 2025-08-19
peports - specify if machine type is unrecognized

wesinator:
followup to #271

Differentiate between an "Unknown" type (null code) and unrecognized machine type in the parsing.


skeeto:
Thanks! I changed the name to "???" because I want a concise, single 
token.

PR #276, opened 2025-09-01
Use ftpmirror.gnu.org for downloads

Peter0x44:
llvm-mingw switched to using it in this commit:

https://github.com/mstorsjo/llvm-mingw/commit/a96bbc0c7fc69e0a7d86678fb3d8c0234d7eb7d2

And I figured we should do the same. Especially since we download more than just GNU make.

GNU themselves write:
We recommend using this generic ftpmirror.gnu.org address wherever possible in links, documentation, etc., to reduce load on the main GNU server. https://www.gnu.org/server/mirror.en.html

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,15 +25,15 @@ RUN apt-get update && apt-get install --yes --no-install-recommends \
 
 RUN curl --insecure --location --remote-name-all --remote-header-name \
     https://downloads.sourceforge.net/project/sevenzip/7-Zip/23.01/7z$Z7_VERSION-src.tar.xz \
-    https://ftp.gnu.org/gnu/binutils/binutils-$BINUTILS_VERSION.tar.xz \
-    https://ftp.gnu.org/gnu/gcc/gcc-$GCC_VERSION/gcc-$GCC_VERSION.tar.xz \
-    https://ftp.gnu.org/gnu/gdb/gdb-$GDB_VERSION.tar.xz \
+    https://ftpmirror.gnu.org/gnu/binutils/binutils-$BINUTILS_VERSION.tar.xz \
+    https://ftpmirror.gnu.org/gnu/gcc/gcc-$GCC_VERSION/gcc-$GCC_VERSION.tar.xz \
+    https://ftpmirror.gnu.org/gnu/gdb/gdb-$GDB_VERSION.tar.xz \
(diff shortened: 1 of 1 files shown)


skeeto:
That's what used to be there, until f8420992. GNU's mirror list included 
too many duds, and about half the time I ran fresh builds, at least one 
GNU download would fail. ftp.gnu.org was far more reliable even with its 
(estimating) mere two 9s of uptime. Though that was 5 years ago; perhaps 
it's worth re-evaluating to see if the mirrors are better now. I'll try it 
out for awhile, and I'm interested in reports of your own experiences.

With Docker caching layers, I don't hit these downloads often, maybe just 
once a month, so I don't have many natural testing opportunities. In #275 
(rudely deleted by @groggily, along with all my comments, which GitHub _so 
stupidly_ allows to happen, making me less likely to reply to randos in 
the future) required adding flex and bison, invalidating the cache, and I 
noticed the GNU servers are very slow right now.

Peter0x44:
Oh, my bad. I guess I should have checked the situation more.
I just saw llvm-mingw doing this and figured we should use it too.

PNBRQK:
>  (rudely deleted by @groggily, along with all my comments, which GitHub _so stupidly_ allows to happen

It seems that his threads is gone with his GitHub account rather than by his hand. Anyway, good news. One junk-maker less

Peter0x44:
> It seems that his threads is gone with his GitHub account rather than by his hand.

If this was the case, then I think his account would simply become "ghost" and all of his threads would still be there.

garoto:
Threads/issues are only removed in their entirety when admins wipe an account from the site for whatever reason. Have seen it in the past more than once. A regular user cannot remove their own issues from the site, only a repository owner can wipe whole issues from their repo issue tracker (have seen that as well).

Peter0x44:
I noticed kernel.org mirrors gnu. Maybe it could be a source?
https://mirrors.kernel.org/gnu/

ghost:
Hi. I'm @groggily. I'm not being banned by Github, nor I deleted my account.

I'm a Proton Mail user. Proton informed me about their `@passinbox.com` feature. I created an alias with `@passinbox.com` and replaced my Github account's email address with it. I don't know that Github silently censored me from the public after that.

Only me can see my contents.

BTW, I'm not junk-maker

skeeto:
I needed a fresh rebuild today, and the recent GNU ftp slowness was 
annoying. Merging this resolved it instantly, so I'm sold!

PR #282, opened 2025-09-10
Don't specify gmp, mpfr, mpc include and lib directories separately

Peter0x44:
I've seen this for a while and wondered why it was done. I ran a build with this and saw no problems.
Maybe some broken configure script was fixed since, I am not sure.


skeeto:
Thanks! These options date to the initial commit, so I probably wrote them 
that way in my original, pre-w64dk scripts. I initially wrote those build 
scripts for GCC 5, which I see has this option even back then, so I'm not 
sure why I wrote it this way. Probably just didn't notice the shorthand.

PR #292, opened 2025-10-06
Add github actions workflow (with caching)

Peter0x44:
This is a continuation of #72, the changes I added were to make caching work, so the CI doesn't have to download tarballs on every run.

~~I am not sure if the changes to the Dockerfile are necessary. I initially made them because exporting the image to layers or something like that was hanging, and it fixed that. Docker is not something I have a lot of knowledge about. My changes were LLM authored, but I did test them as thoroughly as I could.~~

They are not. Removed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,79 @@
+name: Build and test
+
+concurrency:
+  group: w64devkit-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+  pull_request:
+    branches: [ "master" ]
+
+# Must set Settings -> Actions -> General -> Workflow permissions to
+# "Read and write permissions"
+permissions:
+  contents: write
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    env:
+      FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: ["x64", "x86"]
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Apply x86 patch
+        if: matrix.arch == 'x86'
+        run: patch -p1 -i src/variant-x86.patch
+
+      - name: Generate archive name and version string
+        id: filename
+        env:
+          ARCH: ${{ matrix.arch }}
+        run: |
+          GIT_VERSION="$(git describe --tags --always)"
+          echo "version=${GIT_VERSION:1}" >> $GITHUB_OUTPUT
+          echo "archive=w64devkit-$ARCH-${GIT_VERSION:1}" >> $GITHUB_OUTPUT
+
+      - name: Build image
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          tags: w64devkit:${{ matrix.arch }}
+          load: true
+          cache-from: type=gha,scope=w64devkit-${{ matrix.arch }}
+          cache-to: type=gha,scope=w64devkit-${{ matrix.arch }},mode=max
+          
+      - name: Extract artifact
+        run: |
+          docker run --rm w64devkit:${{ matrix.arch }} > w64devkit.exe
+          
+      - name: Rename artifact
+        run: |
+          mv w64devkit.exe ${{ steps.filename.outputs.archive }}.exe
+          ls -lh ${{ steps.filename.outputs.archive }}.exe
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: ${{ steps.filename.outputs.archive }}.exe
+          path: ${{ steps.filename.outputs.archive }}.exe
+          archive: false
+
+      - name: Create Release
+        if: startsWith(github.ref, 'refs/tags/')
+        uses: softprops/action-gh-release@v2
+        with:
+          name: ${{ steps.filename.outputs.version }}
+          files: w64devkit-*.exe
+          draft: true
--- a/Dockerfile
+++ b/Dockerfile
@@ -777,5 +777,5 @@ RUN printf "id ICON \"$PREFIX/src/w64devkit.ico\"" >w64devkit.rc \
  && cat /dl/mingw/mingw-w64-libraries/winpthreads/COPYING \
         >>$PREFIX/COPYING.MinGW-w64-runtime.txt \
  && echo $VERSION >$PREFIX/VERSION.txt \
- && 7z a -mx=9 -mtm=- $PREFIX.7z $PREFIX
-CMD cat /7z/7z.sfx $PREFIX.7z
+ && 7z a -mx=9 -mtm=- w64devkit.7z $PREFIX
+CMD ["cat", "/7z/7z.sfx", "w64devkit.7z"]
--- a/README.md
+++ b/README.md
@@ -31,6 +31,10 @@ This takes about 15 minutes on modern systems. You will need an internet
 connection during the first few minutes of the build. **Note:** Do not use
 PowerShell because it lacks file redirection.
 
+# TEST
+Note that if you want a custom build, you can fork the repo, enable github actions
+push your commits and download the built artifacts.
+
 ## Usage
 
 The self-extracting 7z archive contains tools in a typical unix-like


Peter0x44:
The changes to the Dockerfile seem to be required. I removed them in https://github.com/Peter0x44/w64devkit/actions/runs/18278436300/job/52035735401 and it hung.

Peter0x44:
I managed to remove the changes to the Dockerfile, it still works.

Peter0x44:
OK. There are still a few issues with this.

for PRs, it runs 2x the builds it needs to.
image

this might just be an issue with docker in general, but I had to update one of gcc's patches, but it did not invalidate the cache for x-gcc and continued using the previously cached build.

Peter0x44:
I fixed the duplicate runs on pull requests. Should be good to merge @skeeto .

Peter0x44:
No. I can't get that working.
But at least it cancels the jobs quickly after they start.

Peter0x44:
@skeeto any objections left? I would love to get this in, I'm using it heavily.
It's being used in all of my experimental/research branches.

Peter0x44:
Pinging this again...

skeeto:
Sorry, I've been putting this off because CI/CD is unpleasant for me, and it's difficult to find motivation, especially when the results are unimportant to me. I barely tolerate it even when I'm being paid to deal with it. Embedding a programming language within YAML syntax, and the resulting syntactical nesting, has got to be one of the worst ideas in computer science to gain widespread adoption.

The rename step seems superfluous, but I see that you deliberately added it in order to cache. It seems the rename still isn't actually necessary, and a "breather" step between "build" and 'run" is what really matters? I'm just trying to understand that part.

I looked at your Actions result to see what this all looks like in the end, and I noticed:

> JSON arguments recommended for ENTRYPOINT/CMD

While this issue with signals doesn't matter to our use case, the warning is annoying and it would be nice not to have it. I'm fine with dropping `$PREFIX` from this last step and just hard-coding a name so we can use JSON syntax. Should we do that?

```patch
--- a/Dockerfile
+++ b/Dockerfile
@@ -525,4 +525,3 @@ RUN printf "id ICON \"$PREFIX/src/w64devkit.ico\"" >w64devkit.rc \
  && echo $VERSION >$PREFIX/VERSION.txt \
- && 7z a -mx=9 -mtm=- $PREFIX.7z $PREFIX
-ENV PREFIX=${PREFIX}
-CMD cat /7z/7z.sfx $PREFIX.7z
+ && 7z a -mx=9 -mtm=- w64devkit.7z $PREFIX
+CMD ["cat", "/7z/7z.sfx", "w64devkit.7z"]
```

Dropping that `ENV` removes a layer, too, as a bonus.

Just as a result of making that branch I've got my first Actions result:

https://github.com/skeeto/w64devkit/actions/runs/19975315723/job/57289946957

I haven't set "Read and write permissions" yet, so I expect this to fail. It's still going as I write this. How have you confirmed that the cache works?

The automatic build and insert into a draft on tags seems like it could be convenient, though it depends on how long that typically takes. I can build locally from scratch pretty quickly, and I test my builds before I even push the tag, so generally the delay between pushing the tag and creating the formal release with artifacts is the time it takes to upload the builds plus `source.tar`. If the build comes after the tag, I have to wait, then download and test, and then if something's wrong with the build I've already pushed bad tag. It kind of flips things upside down. I don't think this is solvable, just an inherent limitation.

Is the primary convenience for you having the `.github/` workflow already checked in, instead of you clumsily adding it on top of each of your own changes?

* * *

What follows is a failed attempt to make this easier on myself, and it's not so important to the PR.

It's probably a personal shortcoming, but my frustration increases quadratically on the time between action and feedback, and with CI/CD that means committing, pushing, then waiting minutes on even the slightest change. (One reason I like C so much: It's one of the precious few programming ecosystems where we measure the time between action and result in milliseconds!) I soon found [nektos/act](https://nektosact.com/), which is a neat little utility for running GitHub Actions locally, and would ideally minimize that time. I rebased and squashed your changes to 6ea4c173, then tried it on Windows with:

    $ act --artifact-server-path ~/Downloads

So far most of the time Docker Desktop crashes on the final export step, causing `act` to hang indefinitely, and I have to restart Docker Desktop manually, after which almost no caching occurs and `act` starts over from scratch. It also runs the architecture builds in parallel, so it ends up downloading everything twice, in parallel. (I deliberately organized the Dockerfile so that both architectures can share a download cache, and this isn't taking advantage of it.) One run eventually succeeded and I got these files:

    $ du -sh 1/*/*
    0	    1/skeeto~w64devkit~J4H7Z8.dockerbuild/skeeto~w64devkit~J4H7Z8.dockerbuild.zip
    0	    1/skeeto~w64devkit~WGY9D3.dockerbuild/skeeto~w64devkit~WGY9D3.dockerbuild.zip
    16.7M	1/w64devkit-x64-2.5.0-2-g6ea4c17.exe/w64devkit-x64-2.5.0-2-g6ea4c17.exe.zip
    15.3M	1/w64devkit-x86-2.5.0-2-g6ea4c17.exe/w64devkit-x86-2.5.0-2-g6ea4c17.exe.zip

Obviously they're a little under half the size they're supposed to be. I see a `PK` header, a DOS stub (meaning w64dk *is* in there, uncompressed), and the end of the file has a `PK` table, but otherwise the zip is corrupt. busybox-w32 `unzip` errors with a "short read". It's not simply truncated, as I initially suspect, but half the bytes in the middle have gone missing. I suspect some PowerShell-related bug is processing it as text and mangling it. Looking online I see Windows isn't supported by `act`, and so it's not supposed to work anyway.

Next I try building in my mac, and I immediately get warnings about running on Apple Silicon, and it seems it's not well-supported either. It successfully builds and "uploads" an x64 package, but the x86 package fails (maybe due to a Git error: "failed to identify reference (tag/branch) for the checked-out revision" but there are lots of these?). The x64 package seems to work fine, so a small success. But the cache does not work, and every build is done from scratch. Perhaps caching isn't solvable with `act`.

Peter0x44:
First of all, thanks for getting around to it. I understand it isn't a fun subject and I pretty much agree with your assessment. Yaml is truly horrific. Related: https://ruudvanasseldonk.com/2023/01/11/the-yaml-document-from-hell

But regarding CI itself, I think it's extremely valuable and would do it for any library/executable I make. There is no easier way to get testing on a variety of operating systems and environments. The pain is nearly always worth it, IMO.

> Is the primary convenience for you having the .github/ workflow already checked in, instead of you clumsily adding it on top of each of your own changes?

Yeah, I'm getting tired of lugging that commit around to all of my experiment branches and having to rebase it constantly.
Maybe it's a selfish reason.

> How have you confirmed that the cache works?

painfully. I just push a commit changing something after the toolchain, and observe that it doesn't take as long (~5 minutes), and the log doesn't do as much.

I don't have any evidence to show other than "trust me bro". Unfortunately.

> The rename step seems superfluous, but I see that you deliberately added it in order to cache. It seems the rename still isn't actually necessary, and a "breather" step between "build" and 'run" is what really matters? I'm just trying to understand that part.

I don't remember that being the motivation. I just like having the commit hash in the file name so it's easy to track which commit a specific w64devkit.exe was built from.

> While this issue with signals doesn't matter to our use case, the warning is annoying and it would be nice not to have it. I'm fine with dropping $PREFIX from this last step and just hard-coding a name so we can use JSON syntax. Should we do that?

probably. I don't care about this warning personally, but if it can be fixed without consequence, we should do it.

Peter0x44:
> While this issue with signals doesn't matter to our use case, the warning is annoying and it would be nice not to have it. I'm fine with dropping $PREFIX from this last step and just hard-coding a name so we can use JSON syntax. Should we do that?

I added this change.

Peter0x44:
To prove that the caching works, I added a commit simply changing something in the README.md as you suggested:

> For example, changing a word in README.md should only take a couple of minutes to rebuild. Just the time to recompress the 7z archive with the new README.md, because that file is copied in at the last step. Requiring a full rebuild is locust behavior.

That run took only 5 minutes, the majority of which was spent downloading the cached layers.

Is there anything more I need to do to convince you that it works?

Full build:
image

rebuild:
image

image

Peter0x44:
Pinging this again.

Peter0x44:
Can this get a revisit? I hope I answered all the concerns.

Peter0x44:
https://github.blog/changelog/2026-02-26-github-actions-now-supports-uploading-and-downloading-non-zipped-artifacts/

Github finally offers a way to not zip an artifact. I adjusted this PR to now take advantage of that.

Peter0x44:
the latest restructuring means this PR needed a small rebase.

I also fixed this warning:
```
Warning: Node.js 20 actions are deprecated.
The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, docker/build-push-action@v6, docker/setup-buildx-action@v3.
Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026.
Please check if updated versions of these actions are available that support Node.js 24.
To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file.
Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true.
For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
```
by bumping a few versions of some "actions" it uses.

@skeeto can you please review and merge this?

It's how I just-now obtained my new w64devkit with cmake and ninja!

N-R-K:
Came from https://github.com/skeeto/w64devkit/issues/337#issuecomment-4060668031, I'll add my 2c:

I like the idea of CI, being able to run test/build on hardware/OS I don't have locally available is nice. But practically, nearly every CI I've had the (dis)pleasure of interacting with has been fragile beyond hope. Random container initialization issue, the yaml config format changing, dependencies randomly failing to install and other magic failure that I won't even bother looking into etc.

At this point, building a house of cards would test my patience less than another CI failure that has nothing to do with me or my project.

But in this case in specific, @Peter0x44 seems to be interested in maintaining this since he is getting use of out it (and I can see it being useful to others for testing changes/getting unreleased builds etc). And he's been a regular contributor so he's not some drive-by who will randomly disappear and place burden on the maintainer.

So with that in mind, simply disabling any "ci notification" and punting that maintenance work entirely to peter seems like a viable path. Up to @skeeto to make the call, but that's my opinion on this.

skeeto:
I made some tweaks to the `README.md` changes and merged as ef0ad50bd7c9fd5e78917f10d58202ad63bc9e79. Thanks, @Peter0x44. I followed @N-R-K's tip about disabling CI failure notifications.

Peter0x44:
Thanks!
I will throw up some experimental branches later.

skeeto:
After hesitating for months, it only took a few days for me to find an unanticipated use for Actions. You mentioned ncurses supporting Windows, so I've had Claude working overnight on the problem (tokens are dirt cheap during off-hours this month):

https://github.com/skeeto/w64devkit/compare/master...claude/init-project-setup-TGhNx

The CI job closes the loop and lets Claude see and act upon results, allowing it to work nearly continuously on the problem for 9 hours so far. Sorry for being so stubborn, @Peter0x44!

Peter0x44:
> --disable-tui \

I did a quick review over those changes and noticed this. Are you sure what Claude did there is correct?

Edit:
> Temp: disable TUI to isolate if GDB builds without ncurses at all

OK nevermind I guess it still needs to work...

skeeto:
Other than checking that it's still going and giving it a kick in the middle after Claude Code crashed, I haven't examined any of the work yet. If it "succeeds" and stops but didn't produce the result I wanted, I'll just course correct and set it going again. Unless it's so bad that it's not worth continuing, like disabling all curses support and calling it a success.

Peter0x44:
Just wondering, is there a reason you couldn't just make claude code kick off the docker builds on your own machine?
What is github actions doing exactly that's beneficial?

skeeto:
That would probably work, too, but this is running in an Anthropic cloud instance, entirely disconnected from my own hardware. I had the idea while I was away from my computer, and started it from my phone. I can check in on it from any device.

Peter0x44:
Huh, I did not know that was possible (I have never used claude code).

Hopefully github copilot has some integration like that too. I would be surprised if it doesn't.

Peter0x44:
I noticed the x86 builds are consistently taking much longer. 17m vs 36m.
I wonder if it's beneficial to only run the x86 build if the x64 build succeeds.

Peter0x44:
I can't think of a good reason not to do that, so #349

Peter0x44:
I tried the ncurses build it created, it's functional, but ccmake's issue where the syntax highlighting colors disappear if the terminal is resized still applies.

So at least it didn't fix that. That doesn't mean it can't be an improvement in others ways, of course.

PR #300, opened 2025-11-05
llama.mak: Update for latest (model files, renames, new sources)

Peter0x44:


skeeto:
Thanks again, @Peter0x44!

PR #301, opened 2025-11-07
cppcheck.mak: Update for latest

Peter0x44:


skeeto:
Perfect, thanks! I really appreciate this.

Peter0x44:
I don't actually use cppcheck, I'm just using this to try to figure out gcc's problems with LTO+Os as being a real world "medium" sized codebase with few dependencies I need to worry about. Figured I might as well submit this while I am at it.

(No luck so far unfortunately, just nightmare fuel)

PR #303, opened 2025-11-08
Add download.cpp.o to llama.mak

imweijh:
Adapt to the latest structural adjustments of the llama.cpp program.


skeeto:
Perfect, thanks! These patches are welcome, and I'm happy to get help keeping `llama.mak` up to date. That's what `contrib/` is all about.

PR #306, opened 2025-11-15
Update llama.mak for new version and httplib integration

imweijh:
Using Grok, I got it to work. I don't know what I'm doing.

### DIFF
--- a/contrib/llama.mak
+++ b/contrib/llama.mak
@@ -17,14 +17,18 @@
 #   $ make -j$(nproc) -f path/to/w64devkit/contrib/llama.mak
 #
 # Incremental builds are unsupported, so clean rebuild after pulling. It
-# was last tested at b6980, and an update will inevitably break it.
+# was last tested at b7067, and an update will inevitably break it.
 
 CROSS    =
 CPPFLAGS = -w -O2 -march=x86-64-v3
 LDFLAGS  = -s
 
 .SUFFIXES: .c .cpp .o
-def = -DGGML_USE_CPU -DGGML_VERSION='"0.9.4"' -DGGML_COMMIT='"unknown"'
+def = \
+  -DGGML_COMMIT='"unknown"' \
+  -DGGML_USE_CPU \
+  -DGGML_VERSION='"0.9.4"' \
+  -DLLAMA_USE_HTTPLIB
 inc = \
   -I. \
   -Icommon \
@@ -33,7 +37,8 @@ inc = \
   -Iggml/src/ggml-cpu \
   -Iinclude \
   -Itools/mtmd \
-  -Ivendor
+  -Ivendor \
+  -Ivendor/cpp-httplib
 %.c.o: %.c
 	$(CROSS)gcc -c -o $@ $(inc) $(def) $(CPPFLAGS) $<
 %.cpp.o: %.cpp
@@ -88,6 +93,7 @@ dll = \
   src/llama.cpp.o \
   src/unicode-data.cpp.o \
   src/unicode.cpp.o \
+  src/models/afmoe.cpp.o \
   src/models/apertus.cpp.o \
   src/models/arcee.cpp.o \
   src/models/arctic.cpp.o \
@@ -204,7 +210,8 @@ exe = \
   tools/mtmd/mtmd-audio.cpp.o \
   tools/mtmd/mtmd-helper.cpp.o \
   tools/mtmd/mtmd.cpp.o \
-  tools/server/server.cpp.o
+  tools/server/server.cpp.o \
+  vendor/cpp-httplib/httplib.cpp.o
 
 all: llama.dll llama-server.exe
 


skeeto:
Thanks! I folded `def` since it's getting long. Otherwise this is exactly byte-for-byte what I would have done.

Did you need LLM assistance because you're unfamiliar with `make`? Unfamiliar with building C++? I'm curious about the gap it's filling.

imweijh:
I'm not familiar with `make` or C++ builds. 
Also, will you be updating your evaluations of LLM models? I started learning about LLMs through your articles from last year.
Thanks

skeeto:
> Also, will you be updating your evaluations of LLM models?

I wasn't planning to write a followup since I'm not sure if I have any new insights other than [updating my model list](https://nullprogram.com/blog/2024/11/10/), but your asking is a little push for me to consider it. A quick version of my current thoughts:

Except Mistral Nemo — which I still feel is a unique, special model — I no longer even keep copies of the models I mentioned in my article. I haven't run any, including Mistral Nemo, since early 2025. In large part because I got better hardware, and there's no reason for me to bother with such small models. The other is that some models were supplanted by newer releases (Gemma 3, Qwen 3, Llama 3.3). The only MoE model I listed was Mixtral, which llama.cpp hasn't supported basically since I wrote the article, but MoE has since become the standard for basically any model worth running. Especially with llama.cpp's `--cpu-moe` (note: doesn't apply to the build here).

FIM is a dead end, and nobody bothers training models for FIM anymore. Instead everyone's gone in the direction of agents. Though agents haven't in general done as well as people anticipated they would at this point. I haven't explored agents yet myself because all the options are either propriety (e.g. Claude Code) or implemented too poorly to be considered mature (Python, npm, etc.). In addition to dropping FIM, thinking tokens have been effectively standardized (vs. ad-hoc, ambiguous `<think>`/etc.), and if I wrote Illume today it would be simpler.

Contrary to my expectations, the LLM space has actually narrowed and slowed down in the second half of 2025. I was just barely able to run Qwen3-235B when it came out in mid-2025, and it greatly raised the bar for me. The proverbial nail in the coffin for every other open weights model so far. Then OpenAI released gpt-oss, raising the bar further while being much easier to run, and nothing of note has changed in over three months since. A possible exception might be Kimi K2, but at 1T running it locally is impractical (even for my employer's special on-prem hardware!). So much unlike gpt-oss, which is practical to CPU-inference on regular consumer hardware. When I first got into this hobby last year, things were upending on a weekly basis. Now it's been over three months of running the same model, with nothing else coming close. Slowing down indeed.

In my article I suggested `Q4_K_M` quants, but today you should use `IQ4_XS`. I used to try/keep some `Q8_0` of my favorites, but it's not worth it and I don't bother anymore. `IQ4_XS` is the sweet spot.

So my current advice is that you should run the largest of the two gpt-oss you can manage, ignoring everything else. For me that's 120b, but I expect most people could run 20b. If you have a GPU-enabled llama.cpp (again, not this build) and a discrete GPU, then consider `--cpu-moe`. Always use `--jinja` (unambiguous CoT output, required by some clients). Use the [Unsloth GGUF](https://huggingface.co/unsloth/gpt-oss-120b-GGUF) of the *original* `F16` quantization, because was released already 4-bit quantized (I suspect to prevent abliteration and such). It's a highly-censored model, and it spends thinking tokens checking its internal rules, so there are still legitimate cases for other models. For my uses none of this gets in the way. Perhaps some specialized models do better in their niche (Aya Expanse for translations?).

For closed weights, GPT-5.1 and Sonnet 4.5 are the SoTA, better than anything you can run locally. Gemini and others are not currently at the cutting edge, though not far behind. If gpt-oss doesn't cut it, I try Sonnet 4.5, sometimes with better results. (The OpenAI API [is now so severely locked down](https://old.reddit.com/r/ArtificialInteligence/comments/1mkwowy) as to be practically useless to me, unfortunately.) Despite reports of improvements, I still absolutely have problems with hallucinations. It's the main cause of queries going from time savers to time wasters, after it sends me down the wrong path. (Most recent example for me: Sonnet 4.5 insisting that Swift `Array`s are comparable, and that the resulting compiler errors indicated I had an outdated compiler, which made me waste time tracking that down and trying to update when it turns out that, no, [Sonnet was definitely hallucinating](https://forums.swift.org/t/pitch-making-array-and-friends-conform-to-comparable/78634).)

Maybe I do have enough to say, and so I should just post this...

Peter0x44:
> Though agents haven't in general done as well as people anticipated they would at this point.

Personally, I disagree on this point. Agents are the single point of using LLMs for me, and hooking up an LLM to a shell has been really a huge step up in usefulness to me. I couldn't recommend it more highly.

Regarding local LLMs, I only have a laptop RTX 4060, which being only 8GB VRAM is not enough for gpt-oss 20B.
All the open-weight models I can handle feel like monkeys and not useful for programming. But for other tasks, they certainly perform well. And task-specific tuning makes a big difference.
If I upgraded my machine and tried larger models, maybe I would change my mind.
My employer pays a subscription for me that I believe is $50 a month. And I would totally pay it out of my own pocket, if they didn't.

Ironically, I work on a local LLM runner platform at my job. And I don't really use them personally for programming.
Most of my coworkers are bigger enthusiasts, with 5090 setups, etc.

skeeto:
> I couldn't recommend it more highly.

Interesting. Which agent software do you recommend? I've heard coding agents keep up with expectations, but that otherwise agents struggle to navigate the real world, e.g. an agent to book a complete vacation will need to successfully navigate various crappy websites for hotels/plane tickets/etc. So non-programmers have largely been disappointed with agents so far.

> which being only 8GB VRAM is not enough for gpt-oss 20B

That's why I mentioned `--cpu-moe`, which lets you leverage your GPU despite the "low" VRAM. Due to MoE, that GPU should work well with gpt-oss-20b, or even gpt-oss-120b if you have >64G RAM (to hold the experts). My Windows system has a RTX 3060i, even weaker than your setup (I bought it in mid-2022 and hadn't anticipated the LLM revolution). With CPU inference I get ~10 tok/s, but with a CUDA-enabled llama.cpp and `--cpu-moe` it leverages my GPU, tripling to ~30 tok/s, *and* it runs more quietly (less fan) due to a cooler CPU. If I didn't have an M4 Max (gpt-oss-120b at 60 tok/s), then this would probably be my main LLM setup.

Few models under 30B parameters pass my basic programming evaluations, and so aren't worth consideration in programming, but gpt-oss-20b punches above its weight. I really do think it's suitable as a "first line" model, which is why it's great that it runs locally so well, and sufficient for most LLM uses.

For me, local models isn't about cost — which, as you note, is reasonable for its value — but about privacy and fully owning my development environment (we're here on a w64dk thread after all). If I can run a model myself, then nothing and nobody can ever take it away. No single point of failure.

> And I don't really use them personally for programming.

Yeah, LLMs are utterly useless for me with C programming, which is obviously what I primarily prefer for personal programming, because I fully eclipse them on all axes of capability. They can generate code faster than me, but it takes longer to review and fix that code than to write it myself in the first place. Unlike mentoring, a model does not learn, so spending that time isn't an investment in human capital as it normally would be.

It's occasionally useful for navigating conventional Modern C++, which comes up in my personal programming. They're good at catching typical mistakes, or doing basic, tedious stuff ("Add range-based for loop support to this container."). But even SoTA cannot answer *truly* difficult, and the most interesting, C++ questions — in large part because humans aren't yet sure of some of those answers either! Those answers still must be discovered the old fashioned way.

In general, the less familiar I am with something, the more useful LLMs become because my questions/needs are simpler. Crossing over into my professional programming, they're great for navigating trendy, short-lived ecosystems (e.g. the latest Python bullshit) if it's in the training data. I don't have to waste time learning arbitrary APIs of middling tools/libraries fated to be discarded/redesigned in a few years time. So, yes, definitely more useful professionally than personally, though honestly that's really reflective of the abysmal state — even if cushy and lucrative for its practitioners — of professional software development.

Peter0x44:
> Interesting. Which agent software do you recommend? I've heard coding agents keep up with expectations, but that otherwise agents struggle to navigate the real world, e.g. an agent to book a complete vacation will need to successfully navigate various crappy websites for hotels/plane tickets/etc. So non-programmers have largely been disappointed with agents so far.

Seems to be true, but I haven't tried any of this myself. I think openai released some web browser with an agent?
Either way, I am not booking vacations frequently enough to make this a usecase worth caring about at all.

I remember reading some tweets from anthropic employees talking about how their agent was ordering lunch with doordash automatically every day. I think that's kind of stupid.

I'm using github copilot. But I am not recommending or advertising it, since I haven't looked at alternatives. Some of my coworkers use cursor and say it is better.

Agents still require guidance. But you can see some surprisingly resourceful behavior out of them. I think writing tests is one of their strongest usecases.

> That's why I mentioned --cpu-moe

Thanks. I was unaware, will definitely run some experiments with this.

> Few models under 30B parameters pass my basic programming evaluations, and so aren't worth consideration in programming, but gpt-oss-20b punches above its weight. I really do think it's suitable as a "first line" model, which is why it's great that it runs locally so well, and sufficient for most LLM uses.

Good to hear. Our results are good too. My coworkers wrote a paper about it:
https://arxiv.org/abs/2510.26550

imweijh:
gpt-oss is indeed an excellent model that feels genuinely intelligent.

For tasks related to science, programming, and similar work, I mainly use `gpt-oss 120b` reasoning high, while for humanities and translation tasks, I primarily use `deepseek v3 0324`.

`IBM Granite 4` still seems to focus on FIM. I haven't actually used it myself, as my coding skills aren't sufficient to evaluate it properly.

skeeto:
> `IBM Granite 4` still seems to focus on FIM.

Ah, good point! Even the largest model, "small" (32B-A9B), failed my basic programming eval, so I quickly dismissed it. On revisit I see its FIM works as well as ever, which I'm happy to see. Though [the FIM format slightly changed in 4.0](https://github.com/skeeto/illume/commit/bcb29181) for some reason. (The Unsloth GGUFs correctly list the new tokens, so llama.cpp `/infill` works seamlessly with the new format.)

> My coworkers wrote a paper about it

Interesting, thanks!

Peter0x44:
I ran some tests of CPU only vs cuda with cpu-moe

TPS went from ~15TPS to ~21TPS. Quite a decent speedup.
And TTFT also went from ~7s to ~3s, and depended heavily on the prompt length, where with cpu-moe it varied less.
I would say this is a bigger deal than the faster streaming.

PR #309, opened 2025-11-19
add chat-parser-xml-toolcall.cpp

imweijh:
Generalized XML-style tool-call parsing with streaming support

### DIFF
--- a/contrib/llama.mak
+++ b/contrib/llama.mak
@@ -17,7 +17,7 @@
 #   $ make -j$(nproc) -f path/to/w64devkit/contrib/llama.mak
 #
 # Incremental builds are unsupported, so clean rebuild after pulling. It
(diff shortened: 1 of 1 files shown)


skeeto:
Thanks! If you want to make this a little easier for me then prefix the commit message with `llama.mak: ` as I've been doing before I merge your commits. I like that I can distinguish these changes from the rest of w64dk at a glance in the commit log. This is a popular Git convention, and you can see llama.cpp itself does this to identify the area of a particular commit (metal, server, ggml, etc.).

Peter0x44:
I'm wondering we shouldn't just use `$(wildcard *.cpp)` at this point... this regular breakage is getting a bit much.
Or if you want worse hacks, manually parsing the list out of the CMakeLists.txt somehow.

PR #311, opened 2025-11-24
llama.mak: add model rnd1.cpp

imweijh:
Updated last tested version to b7140


skeeto:
Thanks! Looks like a few more things broke shortly after you opened this, and I just folded it into your commit.

imweijh:
Thanks~

PR #312, opened 2025-11-25
llama.mak: Fix server file references in llama.mak

imweijh:
The order of files seem matter

### DIFF
--- a/contrib/llama.mak
+++ b/contrib/llama.mak
@@ -212,11 +212,11 @@ exe = \
   tools/mtmd/mtmd-audio.cpp.o \
   tools/mtmd/mtmd-helper.cpp.o \
   tools/mtmd/mtmd.cpp.o \
-  tools/server/server-common.cpp.o \
+  tools/server/server.cpp.o \
   tools/server/server-http.cpp.o \
+  tools/server/server-common.cpp.o \
   tools/server/server-queue.cpp.o \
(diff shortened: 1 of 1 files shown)


skeeto:
The order of this listing does not constrain the build order, so this change makes no difference. Most `make` implementations, including GNU Make, build things top-to-bottom because there's no particular reason to use a different order. Sometimes builds accidentally rely on this order, and break in parallel builds, or with `--shuffle` which helps detect such issues.

That's not the case here. These are independent translation units and there is no build order other than linking happening last. If you're having trouble then there's a problem other than order. As noted in `llama.mak`, it does *not* support incremental builds! You must `make clean` on any change and rebuild from scratch, whether due to `git pull` or changes to `llama.mak`. I suspect this is your issue.

There are ways to support incremental builds without much more work (`gcc -MM`), but the point of `llama.mak` is to enable building llama.cpp without any tools beyond w64dk itself (even Git is optional). If you want to regularly rebuild llama.cpp, perhaps to hack on it, just use its CMake build.

skeeto:
Ran it a few times with `--shuffle` and hit a failure, so I think I figured out the issue you're seeing: The dependency on the generated headers moved from `server.cpp` to `server-http.cpp` when it was split out.

imweijh:
Thanks.

PR #313, opened 2025-11-29
llama.mak: support qwen3 next

imweijh:
Updated last tested version from b7150 to b7193.


skeeto:
Perfect, thanks! Merged exactly as-is.

PR #314, opened 2025-12-01
llama.mak: Update last tested version to b7212

imweijh:
Updated last tested version in the makefile.


skeeto:
Thanks! server-models.cpp just popped into existence, so I added that to 
your commit, too.

imweijh:
Thanks~

PR #316, opened 2025-12-04
llama.mak: sync to b7265

imweijh:


imweijh:
Can we use wildcard btw ? like https://github.com/skeeto/w64devkit/pull/309#issuecomment-3555258844

skeeto:
I finally took time to figure out a wildcard solution so that the script is more robust to changes: cfe2d58. Try it out, and assuming no objections, I plan to adopt that instead of your changes here. It will still require changes from time to time, such as 93727df45566154181fb6174e6cd11e6ba978efe, but that should be a lot less frequent.

imweijh:
Thanks~
Using https://github.com/skeeto/w64devkit/commit/cfe2d58143e2ecfca0a8e48041ae0c40ce41583e successfully compiled b7330

PR #320, opened 2026-01-01
llama.mak: remove hardcoded GGML versioning

Peter0x44:
Now that GGML's version has been bumped from 0.9.4 to 0.9.5, the hardcoded version in llama.mak is no longer correct. Instead, parse the version number out of the CMakeLists.txt and retrieve the commit hash using git if available.


skeeto:
Great, thanks! I appreciate that you made it work even without Git.

PR #325, opened 2026-02-06
llama.mak: support building Vulkan backend

Peter0x44:
Requires the Vulkan SDK installed to build.
It can be downloaded from: https://vulkan.lunarg.com/

### DIFF
--- a/contrib/llama.mak
+++ b/contrib/llama.mak
@@ -1,4 +1,4 @@
-# llama.cpp server and DLL build (CPU inference only)
+# llama.cpp server and DLL build (CPU or Vulkan inference)
 #
 # llama.cpp is an amazing project, but its build system is poor and
 # growing worse. It's never properly built llama.dll under any compiler,
@@ -16,12 +16,17 @@
 #
 #   $ make -j$(nproc) -f path/to/w64devkit/contrib/llama.mak
 #
+# To build with Vulkan GPU support, set VULKAN_SDK:
+#
+#   $ make ... VULKAN_SDK=C:/VulkanSDK/1.4.328.1
+#
 # Incremental builds are unsupported, so clean rebuild after pulling. It
 # was last tested at b8124, and an update will inevitably break it.
 
 CROSS    =
 CPPFLAGS = -w -O2 -march=x86-64-v3
 LDFLAGS  = -s
+HOST_CXX = c++
 
 .SUFFIXES: .c .cpp .o
 
@@ -49,10 +54,7 @@ inc = \
   -Itools/mtmd \
   -Ivendor \
   -Ivendor/cpp-httplib
-%.c.o: %.c
-	$(CROSS)gcc -c -o $@ $(inc) $(def) $(CPPFLAGS) $<
-%.cpp.o: %.cpp
-	$(CROSS)g++ -c -o $@ $(inc) $(def) $(CPPFLAGS) $<
+lib =
 
 dll = \
   $(addsuffix .o,$(wildcard \
@@ -65,7 +67,6 @@ dll = \
       src/*.cpp \
       src/models/*.cpp \
    ))
-
 exe = \
   tools/mtmd/clip.cpp.o \
   tools/mtmd/mtmd-audio.cpp.o \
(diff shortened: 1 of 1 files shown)


Peter0x44:
A good talk on what to expect regarding Vulkan in llama.cpp:

https://fosdem.org/2026/schedule/event/CZSPSC-llama-cpp-vulkan/

I don't love depending on the Vulkan SDK, and nothing it does is really irreplaceable, but upstream expects it, so it's what I used.

Peter0x44:
to give a bit more overview of what vulkan sdk provides, just in case something is unclear.

glslc - compiler for glsl -> spir-v. Vulkan does not support text shaders, it instead consumes spir-v IR at runtime.
vulkan headers - self explanatory. In my own projects, I prefer [volk](https://github.com/zeux/volk), which is unofficial.
volk is more-or-less a vulkan equivalent of GLAD. It calls LoadLibrary on `vulkan-1.dll` which ships as part of Windows, at C:\Windows\System32.
(not that it's used here).

vulkan library - vulkan-1.lib import lib for vulkan-1.dll

Though I have not tried to, these components *could* all be built and provided separately, in theory (and likely practice).

Here are the official vulkan build instructions, for reference.
https://github.com/ggml-org/llama.cpp/blob/master/docs/build.md#for-windows-users

skeeto:
Thanks, @Peter0x44. I gave it a test drive and all seems to be in order, including a performance boost on my system when leveraging Vulkan. I'm happy to merge after sorting out a few details:

1. Perhaps instead of `USE_VULKAN` it's activated merely by providing a Vulkan installation, e.g. defining `VULKAN_SDK`. I tested with an older Vulkan SDK I already had on hand, 1.3.280.0. Hard-coding a version like 1.4.328.1 will require maintenance, and users will likely need to provide this anyway. Your hard-coded path would be fine as an example `make` invocation.

2. Instead of `{def,inc,ldflags}_vk` should it `+=` to the appropriate variables? It does so with `dll`. This would require re-ordering a few things.

3. A common subtly in meta-programming builds: Should the shader generator be built with the "cross" compiler? You're already on track to treat it as a different toolchain by not using `CPPFLAGS`. When metaprogramming in a cross-build, there will be two toolchains in play, and it's easy to mix them up. Though it's kind of a moot point here because there's probably no cross Vulkan SDK anyway, so this part isn't cross-build compatible anyway.

4. This is my fault for not enforcing consistency in the past, but my mental model is: `ALL_CAPS` for variables the user might be interested in overriding (possibly by the environment with `make -e`) and `snake_case` for "internal" variables not to be touched by users. `CROSS`, `CPPFLAGS`, and `VULKAN_SDK` are examples of the former, while `def`, `inc`, `dll`, `exe` are examples of the latter. `GGML_VERSION_MAJOR` and such breaks this rule, and I should probably fix it (in a followup to this PR). In your changes some as certainly the latter category, but maybe a couple should be in the former? (Should overriding the shader build directory be a UI knob?)

To check my understanding, I actively made these changes in 770df0307a5cc72a656faf0f816d0baace18e4a7.

Peter0x44:
> Perhaps instead of USE_VULKAN it's activated merely by providing a Vulkan installation, e.g. defining VULKAN_SDK. I tested with an older Vulkan SDK I already had on hand, 1.3.280.0. Hard-coding a version like 1.4.328.1 will require maintenance, and users will likely need to provide this anyway. Your hard-coded path would be fine as an example make invocation.

that's sensible, I was just thinking if anyone wanted to build/supply the components of the vulkan SDK independently.
Fine with me.

> Instead of {def,inc,ldflags}_vk should it += to the appropriate variables? It does so with dll. This would require re-ordering a few things.

Yeah, it can do that. It's sensible

> A common subtly in meta-programming builds: Should the shader generator be built with the "cross" compiler? You're already on track to treat it as a different toolchain by not using CPPFLAGS. When metaprogramming in a cross-build, there will be two toolchains in play, and it's easy to mix them up. Though it's kind of a moot point here because there's probably no cross Vulkan SDK anyway, so this part isn't cross-build compatible anyway.

You are right about this. It should be built with the host compiler.

But, those people will have to still override `glslc` to point to the native one, too, I think.

> This is my fault for not enforcing consistency in the past, but my mental model is: ALL_CAPS for variables the user might be interested in overriding (possibly by the environment with make -e) and snake_case for "internal" variables not to be touched by users. CROSS, CPPFLAGS, and VULKAN_SDK are examples of the former, while def, inc, dll, exe are examples of the latter. GGML_VERSION_MAJOR and such breaks this rule, and I should probably fix it (in a followup to this PR). In your changes some as certainly the latter category, but maybe a couple should be in the former? (Should overriding the shader build directory be a UI knob?)

I think `GLSLC` should also be one of the former. I can't think of a good reason to override the shader build directory, so probably it should be lowercase.

Peter0x44:
Another thing I was going to send a follow up patch for, is openmp. So if we switch this from a `USE_VULKAN`, what would be a sensible pattern for a `USE_OPENMP`?
Not a big deal, just a minor point on consistency.

Peter0x44:
Switch this from:
`  glslc = $(VULKAN_SDK)/Bin/glslc.exe`
to:
`  GLSLC ?= $(VULKAN_SDK)/Bin/glslc.exe`
And update the references to `$(glslc)` and I'd call the patch done.

skeeto:
Update: a8c417bf13d49cfae3707999a28748f884b3c79c (`git diff 770df030..a8c417bf`). I rebased on #330, made the `GLSLC` change, and replaced `vk_shaders_gen_src` with `vk_shaders_gen_obj` to separate compile and link for the shader generator. This allows ccache to cache the generator build, which bottlenecked my cached rebuilds in my testing (I'm quite enjoying ccache). This change no longer lists the headers as a dependency, but that's fine because this build doesn't require that granularity (no incremental builds).

> what would be a sensible pattern for a `USE_OPENMP`?

w64dk will always have OpenMP, so I don't see any reason to turn it off if it's an improvement. It could be this simple:

```patch
diff --git a/contrib/llama.mak b/contrib/llama.mak
index f92e468..47a2200 100644
--- a/contrib/llama.mak
+++ b/contrib/llama.mak
@@ -43,2 +43,3 @@ def = \
   -DGGML_USE_CPU \
+  -DGGML_USE_OPENMP \
   -DGGML_VERSION='"$(GGML_VERSION)"' \
@@ -105,5 +106,5 @@ endif
 %.c.o: %.c
-	$(CROSS)gcc -c -o $@ $(inc) $(def) $(CPPFLAGS) $<
+	$(CROSS)gcc -c -o $@ $(inc) $(def) -fopenmp $(CPPFLAGS) $<
 %.cpp.o: %.cpp
-	$(CROSS)g++ -c -o $@ $(inc) $(def) $(CPPFLAGS) $<
+	$(CROSS)g++ -c -o $@ $(inc) $(def) -fopenmp $(CPPFLAGS) $<
 
@@ -112,6 +113,6 @@ all: llama.dll llama.dll.a llama-server.exe
 llama-server.exe: $(exe) $(dll)
-	$(CROSS)g++ $(LDFLAGS) -o $@ $(exe) $(dll) -lws2_32 $(lib)
+	$(CROSS)g++ -fopenmp $(LDFLAGS) -o $@ $(exe) $(dll) -lws2_32 $(lib)
 
 llama.dll: $(dll) llama.def
-	$(CROSS)g++ -shared $(LDFLAGS) -o $@ $(dll) llama.def $(lib)
+	$(CROSS)g++ -shared -fopenmp $(LDFLAGS) -o $@ $(dll) llama.def $(lib)
 
```

However when I do this it cuts my tok/s in half, so it seems llama.cpp isn't making proper use of OpenMP in w64dk for whatever reason. I also don't see a way to revert to non-OpenMP inferencing if it's compiled in, so I suppose this should be opt-in somehow if it actually improves the results for some people.

skeeto:
f5b06da is the cleanup I intend to put on top.

skeeto:
@Peter0x44, if you're ok with my `llama-vulkan` branch, that's what I'll merge for this PR. I think it's ready.

Peter0x44:
LGTM

Peter0x44:
Thanks!

Peter0x44:
> w64dk will always have OpenMP, so I don't see any reason to turn it off if it's an improvement. It could be this simple:

I wonder what's up with this, since I just checked, the upstream binary builds for windows *are* built with openmp. Maybe it's winpthreads being bad, or something else...

PR #330, opened 2026-02-22
llama.mak: Update last tested version in llama.mak

imweijh:
Updated last tested version from b7761 to b8124.


Peter0x44:
would #325 need this same change?

skeeto:
Thanks!

PR #364, opened 2026-03-28
llama.mak: Update last tested version in llama.mak

imweijh:
The built-in CMake in v2.6 is also very convenient for compilation.


skeeto:
Yup, `llama.mak` isn't so important as it used to be, but its build artifacts are still superior to the CMake build.

PR #365, opened 2026-04-01
llama.mak: Update for latest llama.cpp

Peter0x44:
The HTML, CSS, and JS for the webui got split up into separate files. There's also a new LLAMA_BUILD_WEBUI define required


skeeto:
Thanks!

PR #367, opened 2026-04-14
actions: update actions for Node 24

Peter0x44:
GitHub is going to remove node 20 from Actions:

https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

All the actions versions have been upgraded to the latest major versions available, ahead of the Node 20 deprecation

Move Azure signing from the old Trusted Signing preview action to Artifact Signing v1, update azure/login to v3, and bump actions/download-artifact to v8.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -130,29 +130,29 @@ jobs:
 
     steps:
       - name: Download x64 executable
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: w64devkit-x64-unsigned
 
       - name: Download x86 executable
-        uses: actions/download-artifact@v7
+        uses: actions/download-artifact@v8
         with:
           name: w64devkit-x86-unsigned
 
       - name: Azure login
-        uses: azure/login@v2
+        uses: azure/login@v3
         with:
           client-id: ${{ secrets.AZURE_CLIENT_ID }}
           tenant-id: ${{ secrets.AZURE_TENANT_ID }}
           allow-no-subscriptions: true
 
       - name: Sign executables
-        uses: azure/trusted-signing-action@v0.5.0
+        uses: azure/artifact-signing-action@v1
         with:
           azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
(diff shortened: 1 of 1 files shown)


Peter0x44:
I was curious about the signing stuff.
Apparently, microsoft rebranded it to be able to sign any artifact, not just exes.

https://azure.microsoft.com/en-us/products/artifact-signing

So, I guess the thing I changed is not necessary, but it should still work.

Peter0x44:
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v4, azure/login@v2. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

This was to fix this message on the actions.

dcmake probably needs this too.

skeeto:
All those version are probably what was in Claude's training data, so I'm not surprised it's a little out of date.

I don't think if you push a tag that signing will work. I believe it must be initiated by me. (Otherwise you could sign anything you wanted with my key.) So I pushed a test tag, `t4`, to see how it goes:

https://github.com/skeeto/w64devkit/actions/runs/24398159864

Tags that don't start with `v` are created as drafts, to help with testing CI. So that will create a draft that I believe only you can I can see. If it works, feel free to delete the tag and draft, and ff-merge this.

Peter0x44:
Well, it worked, so no problems there.

Only a minor warning about `actions/cache@0400d5f644dc74513175e3cd8d07132dd4860809` running on node 20.

https://github.com/Azure/artifact-signing-action/issues/125

Apparently, the @v1 tag didn't get updated to point to v1.2.0. I fixed that, now.

PR #377, opened 2026-05-01
Algol68

Peter0x44:
These are my experiments building the new algol68 support of gcc into w64devkit. Not functional.


Peter0x44:
I managed to make it build which is nice. It still can't compile a hello world though...

Peter0x44:
~ $ ga68.exe test.a68
~ $ ./a.exe
Hello, World!

I'm officially the first person to use mingw ga68! 🎉

PR #380, opened 2026-05-10
contrib: Add bootstrap-go-toolchain.sh

Peter0x44:
Ever since w64devkit began forcing bigobj by default, it has been incompatible with cgo. I patched go to support bigobj, but unfortunately upstream the patch ran in to bureaucracy/difficulties and I ran out of patience to get it through. So that means it's not really an option to use upstream go binaries, it must build from source. Bootstrapping go is not difficult, but it has grown more tedious as more required stages have been added to the bootstrap chain.

So, to spare myself some tedium, I have created
bootstrap-go-toolchain.sh. It builds the latest version of go, from source, using only w64devkit's gcc compiler. It also adds my bigobj patches to the toolchain, so cgo works properly.

It might deserve a mention on:
https://nullprogram.com/blog/2020/01/21/

### DIFF
--- a/contrib/gobuild/bootstrap-go-toolchain.sh
+++ b/contrib/gobuild/bootstrap-go-toolchain.sh
@@ -1,0 +1,151 @@
+#!/bin/sh
+# Build Go from source on Windows using w64devkit.
+# Bootstraps from Go 1.4 through intermediate versions up to the latest.
+#
+# Usage: ./bootstrap-go-toolchain.sh [INSTALL_DIR]
+#   INSTALL_DIR  Where to install the final Go toolchain (default: ./go)
+#
+# Prerequisites: w64devkit in PATH (provides gcc, wget, tar, patch, sh)
+#
+# The final version is patched with go-bigobj.patch located next to
+# this script before building.
+# Tarballs are verified against the SHA-256 values embedded below.
+#
+# Bootstrap chain (per https://go.dev/doc/install/source):
+#   Go 1.4  -> built with C compiler (gcc from w64devkit)
+#   Go 1.17 -> built with Go 1.4
+#   Go 1.20 -> built with Go 1.17
+#   Go 1.22 -> built with Go 1.20
+#   Go 1.24 -> built with Go 1.22
+#   Go 1.26 -> built with Go 1.24
+#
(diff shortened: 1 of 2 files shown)


skeeto:
Tried it out and it worked, thanks! It's nice to have cgo working again. I like that it caches the downloads. If it didn't I was going to suggest it.

Except a small hiccup: I had to unset `GOBIN`. There's a long-standing bug in the Go Windows bootstrap such that it breaks when this variable is present. I suggest unsetting it in the script. Other than that I'm happy with this.

Peter0x44:
> Except a small hiccup: I had to unset GOBIN. There's a long-standing bug in the Go Windows bootstrap such that it breaks when this variable is present. I suggest unsetting it in the script. Other than that I'm happy with this.

Done.

skeeto:
> It might deserve a mention on:  
> https://nullprogram.com/blog/2020/01/21/

What mention did you have in mind? I thought it would be something about w64dk requiring this bootstrap, but on revisit I remember now that this article predates w64dk.

Peter0x44:
I can extend it to support Linux too. Then mention that this script can build a toolchain for you.

Peter0x44:
It doesn't really mention that now go has added extra stages to the bootstrap as well, so you have to go through several versions.

Peter0x44:
It works for Linux now too. I would also add mac, it's probably not much work, but I have no way to test it.

skeeto:
This is enough to get it started on macOS Apple Silicon:
```diff
--- a/contrib/gobuild/bootstrap-go-toolchain.sh
+++ b/contrib/gobuild/bootstrap-go-toolchain.sh
@@ -32,3 +32,3 @@ HOST_OS="$(uname -s)"
 case "$HOST_OS" in
-    Linux*)
+    Linux*|Darwin)
         MAKE_COMMAND=./make.bash
@@ -81,3 +81,3 @@ verify_download() {
         cd "$DOWNLOADS"
-        printf '%s\n' "$entry" | sha256sum -c
+        printf '%s\n' "$entry" | sha256sum -c -
     )
```
But it immediately gets stuck on Go 1.4 because Go did not yet support macOS on ARM. It was only 2014 after all. I expect bootstrapping is closed to Apple Silicon.

Peter0x44:
Oh, then it is not worth bothering.

PR #392, opened 2026-05-21
[PROPOSAL] make2compdb.c

### LINKED ISSUE #251: Speculative application: make2compdb
In #246 I came up with an idea to generate a [JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) from GNU Make output. This concept has been realized before, and I found [two existing, if crude, implementations](https://stackoverflow.com/questions/21134120/how-to-turn-makefile-into-json-compilation-database). I came up with the name `make2compdb`, which is the name I'll use here, but I'm open to other names. In it simplest usage:

    $ make -Bnw | make2compdb >compile_commands.json

The user can invoke `make` in whatever way is appropriate, choosing targets, overriding `CFLAGS`, etc. The three `make` options are necessary:

* `-B` (`--always-make`): force all build commands to print regardless of current build state

* `-n` (`--dry-run`): no build is actually desired at this time (this still invokes `!=` and `$(shell ...)`, which is a good thing)

* `-w` (`--print-directory`): The `"directory"` key in a compilation database is required, and despite what the "specification" says must be an absolute path. At least some LLVM tools (e.g. `clang-tidy`) do not support relative paths. This is highly unfortunate because it means `compile_commands.json` cannot be meaningfully checked into source control. `-w` provides the `"directory"` key for `make2compdb` via "Entering directory" lines, which internally maintains a stack of directories. The stack is initialized with the absolute working directory.

The output of `make` will look something like:

    make: Entering directory 'C:/Users/me/src/example'
    cc -c -Iinclude -o example.o example.c
    c++ -c -Iinclude -o util.o util.cpp
    make[1]: Entering directory 'C:/Users/me/src/example/lib'
    cc -c -I../include -o lib.o lib.c
    make[1]: Leaving directory 'C:/Users/me/src/example/lib'
    c++ -o example.exe example.o util.o src/lib.o
    make: Leaving directory 'C:/Users/me/src/example'

From this, the output of `make2compdb` would be (assumed UTF-8):

```json
[
    {
        "directory": "C:/Users/me/src/example",
        "file": "example.c",
        "command": "cc -c -Iinclude -o example.o example.c"
    }, {
        "directory": "C:/Users/me/src/example",
        "file": "util.cpp",
        "command": "c++ -c -Iinclude -o util.o util.cpp"
    }, {
        "directory": "C:/Users/me/src/example/lib",
        "file": "lib.c",
        "command": "cc -c -I../include -o lib.o lib.c"
    }
]
```

The program does not need to access nor examine any of these files, but transform data from one format to another. This should be easy, but since we're bothering to write a whole program to do it, it ought to be robust too. The `"directory"` fields are copied verbatim from `'...'` in the `Entering` line, but the rest must be parsed the same way [as would a shell](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/V3_chap02.html), and with some understanding of what it sees (note the distinction between "must" and "should"):

* Must ignore lines that don't look like a compiler invocation. It should match `cc`, `c++`, `gcc`, `g++`, `clang`, `clang++`, and even fancier stuff like `x86_64-w64-mingw32-gcc`, which can be done by splitting on `-` and looking at the last element. It should must absolute paths, too, like `C:/w64devkit/bin/gcc.exe`. If you want to be fancy, handle `cl` and `clang-cl`, too, but specially.

* It must handle quoting and backslashes when word splitting:

        "C:/Users/John Falstaff/w64devkit/bin/gcc.exe" -c ...
        C:/Users/John\ Falstaff/w64devkit/bin/gcc.exe -c ...
        C:/Users/'John Falstaff'/w64devkit/bin/gcc.exe -c ...

* It must handle line continuations:

        gcc \
          -c ...

  The continuation backslash and newline should probably be omitted from `"command"`. I don't know how well database-consuming tools handle this.

* It should navigate around subshell expressions:

        gcc -c $(pkg-config ...) ...

  That is, ignore the stuff between `$(...)`. In the case of `pkg-config`, a good Makefile will expand these on its own first — which is great, as the output will end up in the compilation database — and they won't appear as subshell expressions in make's output. So trying to actually interpret them is unimportant.

* It should understand common compiler flags like `-o`, `-I`, etc., so `-o foo.c` will not match `foo.c`. It should only match known file extensions.

* If a command contains multiple matching positional arguments, there must be a compilation database for each. `cc x.c y.c` will produce an entry for `x.c` and `y.c`.

* Optional (bonus) variable expansion:

        "$CC" -c ...

  This is not so easy (consider variations like `${...}`). Typically these are handled as make variables, and are already expanded on output, so this is low importance. Probably better to not do this at all than do it poorly.

This all assumes a unix shell (e.g. `sh.exe`), but some makefiles are [written for `cmd.exe`](https://peter0x44.github.io/posts/make_shell_cmd/), and perhaps via an option can be instructed to parse like `cmd.exe` instead. Output should be properly-formatted JSON, which is easy. Output that cannot be understood, or which cannot be represented as JSON, is silently ignored.

The usual utility requirements stand:

* Must support Windows XP.

* Must not link MSVCRT (e.g. `cc -nostartfiles -o make2compdb make2compdb.c -lmemory`).

* Must be a single source file, like `make2compdb.c`. C++ is acceptable, but mind the previous item.

* Must be dedicated to the public domain. Still put your name in it and take credit of course.

As mentioned, I expect this is a pretty easy project. Inputs and outputs are relatively small, and must fit in memory in practical use. So it should slurp the entire input before starting. That means as far as system interfaces goes, it only needs to read from standard input, write (buffered!) to standard output, and retrieve the working directory (to seed the directory stack).

Looking at the various tools that consume compilation databases, I do not see a reason I'd personally use any of them. So I'm not really the right person to build this tool, as I wouldn't be using it. So I'm opening it up to anyone who would like to explore the idea.

g-berthiaume:
This PR introduces `make2compdb`.
A new CLI tool to generates Clang's JSON Compilation Database files (`compiler_commands.json`) from make build systems.

The API can be used in the following way:

```sh
$ make -Bwn | make2compdb.exe > compiler_commands.json
$ cat compiler_commands.json
[
  {
    "directory": "C:\\my_project",
    "file": "main.c",
    "output": "main",
    "arguments": [
      "gcc",
      "-o",
      "main",
      "main.c"
    ]
  }
]
```

### Properties

- Single C23 source file
- Compiles on both Windows and Linux
- Unit tests
- Basic fuzzing (I'm pretty new to this)
- CRT-free on Windows 
- Acceptably fast

<details>
<summary>Expand to see peports</summary>

```sh
$ peports make2compdb.exe
KERNEL32.dll
        0       ExitProcess
        0       GetCommandLineW
        0       GetConsoleMode
        0       GetCurrentDirectoryW
        0       GetStdHandle
        0       ReadFile
        0       VirtualAlloc
        0       WriteConsoleW
        0       WriteFile
SHELL32.dll
        0       CommandLineToArgvW
```

</details>

<details>
<summary>Expand to see performance benchmarks</summary>

While I'm sure, we could do better, the CLI tool seems to be acceptably fast.
It seems to be a bit faster on Linux than on Windows, but I'm not sure if I'm not just measuring piping speeds.

With a small project (linux)

```sh
$ hyperfine "./make2compdb < make_output.txt" --warmup 5
Benchmark 1: ./make2compdb < make_output.txt
  Time (mean ± σ):       5.8 ms ±   1.0 ms    [User: 1.1 ms, System: 0.9 ms]
  Range (min … max):     4.1 ms …  11.9 ms    321 runs
```

With a small project (windows)

```sh
$ hyperfine ".\make2compdb.exe < make_output.txt" --warmup 5
Benchmark 1: .\make2compdb.exe < make_output.txt
  Time (mean ± σ):      13.5 ms ±   1.0 ms    [User: 8.6 ms, System: 7.2 ms]
  Range (min … max):    12.3 ms …  18.1 ms    112 runs
```

With ffmpeg (windows)

```sh
$ hyperfine "make2compdb.exe < ffmpeg_output.txt" --warmup 5
Benchmark 1: make2compdb.exe < ffmpeg_output.txt
  Time (mean ± σ):      55.8 ms ±   4.5 ms    [User: 29.0 ms, System: 25.9 ms]
  Range (min … max):    51.4 ms …  79.0 ms    45 runs
```

With ffmpeg (linux)

```sh
$ hyperfine "make2compdb < ffmpeg_output.txt" --warmup 5 
Benchmark 1: make2compdb < ffmpeg_output.txt
  Time (mean ± σ):      28.2 ms ±   0.6 ms    [User: 22.8 ms, System: 3.7 ms]
  Range (min … max):    27.2 ms …  30.3 ms    102 runs
```

By comparison, the https://github.com/nickdiego/compiledb takes 18 seconds to analyzer ffmpeg.

```sh
$ hyperfine "compiledb -p ffmpeg_output.txt"
Benchmark 1: compiledb -p ffmpeg_output.txt
  Time (mean ± σ):     18.469 s ±  0.258 s    [User: 18.128 s, System: 0.205 s]                                        
  Range (min … max):   18.127 s … 19.003 s    10 runs
```

So it's a 331x speedup.

</details>

### Features

- Support makefiles using gcc, clang, their variants (clang-22, arm-none-eabi-gcc, etc) and zig cc.
- Support Windows path (`\\`), EOL (`\r\n`) and executable extension (`.exe`).
- support multi source files build command (e.g. `gcc main.c lib.c -o main`).
- Support single header library
  Address this issue https://github.com/nickdiego/compiledb/issues/140
- support different make name (remake, bmake, etc)
  Address this issue https://github.com/nickdiego/compiledb/issues/146
- support special make quotation marks
  Address this issue https://github.com/nickdiego/compiledb/issues/102
- support compiler wrapper (ccache, distcc, etc)
  Address this issue https://github.com/nickdiego/compiledb/issues/2

#### 1. No Make mode

I'm a big fan of the "unity build" compilation technique (like the one used in u-config).
Therefore, I often don't have a use for a makefiles in my projects.
That said, I still need `compiler_commands.json` for my IDE to work.

How great would it be to just pipe my build command into `make2compdb` ?
Well you can!

```sh
$ echo "gcc main.c -o main" | make2compdb.exe
[
  {
    "directory": "C:\\my_project",
    "file": "main.c",
    "output": "main",
    "arguments": [
      "gcc",
      "-o",
      "main",
      "main.c"
    ]
  }
]
```

#### 2. `--verbose`

I believe that there's a lot of value in providing your technical users tools to help them self-diagnose issues.
In my experience, this usually leads to better bug reports and therefore help maintaining the project.

This is why I have added the `--verbose` CLI flag.
When it's passed to `make2compdb`, the stdout contains debug information.

```sh
$ make -Bwn | make2compdb.exe --verbose
make2compdb
Version: 2026-05-20
Verbose mode: true
Directory: "C:\\my_project"
CLI args: ["./make2compdb", "--verbose"]

------
Step 1: Identifying the parsing mode
    Parsing mode is SHELL

Step 2: Parsing shell command
    Invocation
        Input: "gcc -o main main.c\n"
        Tokens: ["gcc", "-o", "main", "main.c"]
        Compiler: {
            .ok = true
            .source = ["main.c"]
            .output = "main"
            .args = ["gcc", "-o", "main"]
        }
[...]
```

### Limitations

1. It does not support Microsoft CL.exe. I would be open to adding this feature if there's demand for it.
2. On Linux, we only support C.UTF-8 and other English UTF8 locales.
   As a non-native English speaker, I can empathize with this being a pain point.
   That said, I didn't find a robust way to parse the Makefile output in different languages.

--- 

### Note for reviewer

I've learn a lot building this project: Arena, fuzzing, CRT-less windows programming.

When starting this project, I think I underestimated the number of corner cases to handle.
To be honest, the shell parsing humbled me a little bit. :^)

One of the reasons I appreciate w64devkit, is its high standard when it comes to programming.
I think I have a lot to learn from the maintainers of this project, so I welcome any of your suggestions on how to improve `make2compdb`.

Closes https://github.com/skeeto/w64devkit/issues/251


Peter0x44:
Epic! I'll be test driving this myself.

skeeto:
Thanks so much, @g-berthiaume! It's interesting to see these concepts through another person's lens.

Make all functions except the entry point `static`. The program is a single translation unit, which communicates to the compiler that functions do not need external linkage. For example, it should inline any function, regardless of size, called only from a single location as there's no downside to doing so. But external linkage counts as some unknown number of other call sites. IMHO, C gets this inverted: `static` ought to be the default, and external linkage ought to be opt-in (via a declared interface).

When I compile with GCC 16 in the latest w64dk I get a `-Wnonnull-compare` (via `-Wall`) warning on line 769, asserting that a `[static]` parameter is non-null. This is essentially GCC warning that it's going to eliminate the assertion because UB must have occurred before it could fire. That means this is not a well-placed assertion.

It's "peek" not "peak". I was genuinely confused by this because the latter would normally mean highest value (highest memory use, highest count, etc.). Also "threated" instead of "treated" and "stars" instead of "starts", and "emtpy" instead of "empty".

The program always outputs UTF-16, and there seems to be a misunderstanding how output is supposed to work. I'm surprised the resulting `compile_commands.json` actually works with anything (did it?): 

    $ printf 'cc x.c' | make2compdb | xxd | head -n1
    00000000: 5b00 0a00 2000 2000 2000 2000 7b00 0a00  [... . . . .{...

If the output device is a console, use `WriteConsoleW` with UTF-16. Otherwise use WriteFile with UTF-8. Casting `c16 *` to `u8 *` does not make it UTF-8. The program is already UTF-8 most of the way through, and on Windows the output buffer is UTF-16. The common case is writing to a JSON file, not to a console, and so this means the common case would convert UTF-8-to-UTF-16 then UTF-16-to-UTF-8, which is silly. Instead use a UTF-8 buffer and, if a console is detected, convert UTF-8-to-UTF-16 at the last moment (caveat: tricky edge case around straddlers which even Microsoft's CRTs get wrong). It's more important that it outputs UTF-8 than that it prints UTF-16 to the console correctly, which is mostly a bonus to make visual inspection/debugging more reliable.

As a see it for yourself test of the above, these must both produce the same result:

    $ printf 'cc π.c' | make2compdb | grep file | xxd
    00000000: 2020 2020 2020 2020 2266 696c 6522 3a20          "file": 
    00000010: 22cf 802e 6322 2c0a                      "...c",.

    $ printf 'cc π.c' | make2compdb >json && grep file json | xxd
    00000000: 2020 2020 2020 2020 2266 696c 6522 3a20          "file": 
    00000010: 22cf 802e 6322 2c0a                      "...c",.

The parser should be more oriented around tokens than substring search. Don't substring search for `"clang"`, match the `"clang"` token. You can further tokenize tokens as part of this process, first tokenizing to `"x86_64-w64-mingw32-gcc"`, then to `"x86_64" "w64" "mingw32" "gcc"` to match `"gcc"`.

Because it's substring matching, this doesn't work as intended:

    $ printf 'ccache cc -c x.c' | make2compdb

Produces an `arguments` of `["ccache", "cc", "-c", "x.c"]` because `ccache` matches `cc`, whereas, say, `mycache` works fine. Add a test for this when you fix it.

`print_str_escaped_string` produces invalid JSON for some inputs:

    $ printf "cc -I'a b' x.c" | make2compdb

Produces output with escaped single quotes, `"-I\'a b\'"`. JSON output could use some unit tests.

The `PARSING_MODE_SHELL` loop doesn't seem to accumulate commands, just keep the last, so:

    $ printf 'cc a.c; cc b.c' | make2compdb

Only captures the second command. Add a test for this when you fix it. Unless this is intentional?

The `&>>` branch doesn't peek correctly, though it works out by chance anyway.

Compiler `--output` flag isn't handled, though there's an aspirational comment about it.

`is_source_file` should probably also cover at least `.s` and `.S` (assembly).

`enum u32` is a strange name. Did you mean `enum : u32`?

Trivial to adjust, but even 16MiB arena may be a little tight for larger builds. Maybe 64M? Build commands can be quite bloated.

* * *

Nothing below is a problem, nor must change. Just pointing them out as commentary.

If you're committed to GNU-style toolchains (e.g. GCC, Clang), as MSVC is excluded, then because you're linking `-lmemory`, `__builtin_memset` (and `__builtin_memcmp`, etc.) is a first-class feature. In release builds, compilers likely figure out `memory_set` on their own, which is what makes `-lmemory` necessary, but it's nice to have a fast memory clear primitive even in debug builds.

Might be surprising (or might not given those `const *const`), but this is a semantically valid function:

```c
void example(Str const s)
{
    s.ptr[0] = 0;
}
```

But `s` is `const`, right? Sure, but it doesn't propagate to pointed at objects. So the `const` is doing practically nothing. That `const` is so pathetic in C and C++ is why I don't bother with it in code I write. Though perhaps you feel it's still useful as documentation. On the other hand, it's not used consistently.

Peter0x44:
> On Linux, we only support C.UTF-8 and other English UTF8 locales.
> As a non-native English speaker, I can empathize with this being a pain point.
> That said, I didn't find a robust way to parse the Makefile output in different languages.

Would calling setlocale before starting the process not help? Then it wouldn't matter what the user's locale is.

g-berthiaume:
@skeeto Thank you for your review.
I agree with every point you mentioned. I will correct those problems later this week.

> @Peter0x44 
> Would calling setlocale before starting the process not help?

I'm not a Linux expert, but even if we assume that the shell starts `make` and `make2compdb` at exactly the same time, I'm not sure we could reliably set the locale without creating a locale "race condition". I'm not even sure when this global variable is read... is it at every write, or maybe in the crt0 before the main?

That said, even if we could, I would prefer not to touch the locale config. In cases of a crash, we could break users environment because `make2compdb` has set the locales before exiting. 
What do you think?


---
EDIT (2026-05-29): 
I have decided to rewrite my shell parser to make it cleaner and more robust. 
My update will be delayed to next week.

g-berthiaume:
Round 2!

The focus for this new release has been to make `make2compdb` more token-oriented.

### Changes

- All functions except the entry point are `static`.
- Fix: All targets now compile without warnings `-Wall -Wextra -Wpedantic -Wconversion`.
- Fix: Remove English typos like `peak`-> `peek` (sorry for the confusion).
- Fix: Windows now uses the correct encoding when outputting to pipes/files.

As you can see, the output is now identical, regardless of the destination.

```sh
λ printf 'cc π.c' | ./make2compdb | grep file | xxd
00000000: 2020 2020 2020 2020 2266 696c 6522 3a20          "file":
00000010: 22cf 802e 6322 2c0a                      "...c",.

λ printf 'cc π.c' | ./make2compdb >json && grep file json | xxd
00000000: 2020 2020 2020 2020 2266 696c 6522 3a20          "file":
00000010: 22cf 802e 6322 2c0a                      "...c",.
```

- Improve the compiler detection. The parser is now more token-oriented (more on that later).

Now `ccache` won't be detected as `cc`.

```sh
λ printf 'ccache clang -c x.c' | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "x.c",
        "arguments": [
            "clang",
            "-c",
            "x.c"
        ]
    }
]
```

- Rework string shell unescaping and json escaping to be compliant with the JSON Compilation Database Format Specification.

```sh
λ printf "cc -I'a b' x.c" | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "x.c",
        "arguments": [
            "cc",
            "-Ia b",
            "x.c"
        ]
    }
]
```

- Fix: Detect comma-separated commands

```sh
λ printf 'cc a.c; cc b.c' | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "a.c",
        "arguments": [
            "cc",
            "a.c"
        ]
    },
    {
        "directory": "C:\\dev",
        "file": "b.c",
        "arguments": [
            "cc",
            "b.c"
        ]
    }
]
```

- Fix: `&>>` redirect operator is now working properly

```sh
λ printf 'gcc main.c -o main &>>test.txt' | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "main.c",
        "output": "main",
        "arguments": [
            "gcc",
            "-o",
            "main",
            "main.c"
        ]
    }
]
```

- Fix: Now support the longer `-o` flag versions: `--output` and `--output=`.

```sh
λ printf 'gcc main.c --output main' | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "main.c",
        "output": "main",
        "arguments": [
            "gcc",
            "--output",
            "main",
            "main.c"
        ]
    }
]
```

- Fix: `str_is_source_file` now includes `.s` and `.S` (assembly).
- Fix: enum `enum u32` -> `enum : u32`.
- Increase memory arena size from 16MiB to 64MiB.
- Now use `__builtin_memset` and `__builtin_memcpy` (provided by `-lmemory`).
- Standardize `const` usage.
- Fix: You can now use more than 1 compiler wrapper

```sh
λ printf "distcc ccache clang main.c" | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "main.c",
        "arguments": [
            "clang",
            "main.c"
        ]
    }
]
```

- Fix: The `-x none` flag now correctly resets the language.

Notice that `some_include.h` is not treated as a source file.

```sh
λ printf "gcc -x c lib.h main.c -x none some_include.h" | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "lib.h",
        "arguments": [
            "gcc",
            "-x",
            "c",
            "lib.h",
            "main.c",
            "-x",
            "none",
            "some_include.h"
        ]
    },
    {
        "directory": "C:\\dev",
        "file": "main.c",
        "arguments": [
            "gcc",
            "-x",
            "c",
            "lib.h",
            "main.c",
            "-x",
            "none",
            "some_include.h"
        ]
    }
]
```

- Fix: Shell redirections can now be used inside the compiler invocation (not just the edges).

Here's a (nasty) example:

```sh
λ printf 'gcc main.c -o 2&> lib.c main' | ./make2compdb
[
    {
        "directory": "C:\\dev",
        "file": "main.c",
        "output": "main",
        "arguments": [
            "gcc",
            "-o",
            "main",
            "main.c"
        ]
    }
]
```

- Rework shell tokenizer.


### Details

At its core, this release focuses on the shell tokenizer re-implementation.
With a better tokenizer, it's easier for the rest of the code to be more token-oriented.
The shell should now be more robust, more tested and more readable.
Also, the tokenizer is now allocation-free, which is a nice bonus.

After thinking about it and re-reading [Arenas and the almighty concatenation operator](https://nullprogram.com/blog/2024/05/25/), I realized that I could simplify my memory allocation strategy by composing `str_concat` and `str_duplicate`. I'm pretty happy with the results.

Performance-wise, the new parsing strategy is a bit slower, which is to be expected, as we now look at every token.
On my machine, ffmpeg is parsed in under 80ms on both Linux and Windows, which is still acceptable.

### Note for reviewer

Thanks again for your last feedback.
Please let me know if you see any areas where this tool could be improved.

g-berthiaume:
@skeeto @Peter0x44 
Just a small heads up: 
I'll be on vacation next week, and therefore away from my keyboard. 
I'll be back on the 21.

src/make2compdb.c
```diff
@@ -0,0 +1,3471 @@
...
+    while (!end_of_line && (shell_input->len > 0)) {
+        Str token = shell_tokenize_next(shell_input, &end_of_line);
+        if (token.len > 0) {
+            strlist_push_back(&tokens, perm, token);
+        }
+    }
+    return tokens;
+}
+
+// :: Compiler
```
Peter0x44, line 1475: What's the purpose of this CompilerKind? It doesn't seem to be doing anything needed?

g-berthiaume: You're right, currently `CompilerKind` is only used in the `CompilerInvocation` struct, which is not really needed.

The idea is that I could extend `make2compdb` to support other compilers like Microsoft's `CL.exe`.

You can see the scaffolding for executing this idea in `compiler_command_from_invocation()`:
image

sleeptightAnsiC: I don't believe this distinction between compilers will be ever needed.

The chance for some flag to clash when, e.g. one compiler expects it to consume the argument while the other does not, it's probably not something worth worrying about.

GCC, Clang and all it's forks (emcc, zig cc, icx, etc...) they all share flags which should be compatible. For CL the flags are usually prefixed with slash `/` rather than dash `-`, so you can cheat a bit here :)

As I mentioned in https://github.com/skeeto/w64devkit/issues/251#issuecomment-4364942668, hardcoding compilers/flags will never be reliable, so you may as well ignore small incompatibilities and keep the code simple, that is, just detect if the compiler is known, but don't try to do anything smart about it.

src/make2compdb.c
```diff
@@ -0,0 +1,3471 @@
...
+            // NOTE: On directory parsing.
+            // Because of the `-w` flag, `make` will print information everytime it leaves a directory.
+            // Something like "make: Leaving directory 'C:/dev/'"
+
+            Str dir = directory_from_make_dir_line(&input);
+            if (dir.len > 0) {
+                Arena scratch = *perm;
+
+                Str cwd = dirstack_pop(&dir_stack, &scratch);
+                (void)cwd;
```
sleeptightAnsiC: The "gymnastic" around returning cwd is a bit odd. `dirstack_pop` is only used in this one place. Wouldn't it make more sense to just edit it like so [diff below] ?

tbh, the code related to dirstack_* and strlist_* seems like (potentially) unnecessary boiler plate, but I haven't looked through all of it though.

```diff
diff --git a/src/make2compdb.c b/src/make2compdb.c
index 39f9d27..46ce8a0 100644
--- a/src/make2compdb.c
+++ b/src/make2compdb.c
@@ -838,15 +838,12 @@ static Str dirstack_peek(DirectoryStack ds, Arena *perm)
     return directory_copy;
 }
 
-static Str dirstack_pop(DirectoryStack *ds, Arena *perm)
+static void dirstack_pop(DirectoryStack *ds)
 {
     Str directory = strlist_pop_back(&ds->stack);
     if (directory.len > 0) {
-        Str directory_copy = str_duplicate(perm, directory);
         arena_reset_to(&ds->arena, (byte *)directory.ptr);
-        return directory_copy;
     }
-    return directory;
 }
 
 // :: OS
@@ -2152,10 +2149,7 @@ static int make2compdb(Arena *perm, OsWriterInterface *w, StrList cli_args, Str
 
             Str dir = directory_from_make_dir_line(&input);
             if (dir.len > 0) {
-                Arena scratch = *perm;
-
-                Str cwd = dirstack_pop(&dir_stack, &scratch);
-                (void)cwd;
+                dirstack_pop(&dir_stack);
             }
             break;
         }
```

src/make2compdb.c
```diff
@@ -0,0 +1,3471 @@
...
+// Main program function.
+static int make2compdb(Arena *perm, OsWriterInterface *w, StrList cli_args, Str make_stdout, Str initial_working_directory)
+{
+    Str const help_message = SL("make2compdb - Create a compile_commands.json from a Make output\n"
+                                "Usage: make2compdb [option]\n"
+                                "\n"
+                                "Options:\n"
+                                "       --version:  Display the version number.\n"
+                                "   -h, --help:     Display this help page.\n"
+                                "   -v, --verbose:  Display debug information to help with problem diagnosis.\n"
```
sleeptightAnsiC: Any chance for diagnostic to appear at `stderr` when using --verbose and similar, so it can be easily piped and split from json output?

I see there is some platform specific separation of stdin/out/err already, but the diagnostic still goes to stdout, so it's not really practical to deal with. (basically, everything that isn't json should go to stderr)

src/make2compdb.c
```diff
@@ -0,0 +1,3471 @@
...
+            SL("-idirafter"),
+            SL("-isysroot"),
+            SL("-imultilib"),
+            SL("-Xpreprocessor"),
+            SL("-Xassembler"),
+            SL("-Xlinker"),
+            SL("-bundle_loader"),
+            SL("--param"),
+            SL("-arch"),
+            SL("-wrapper"), SL("-wrapper"),
```
sleeptightAnsiC: `-wrapper` twice?

g-berthiaume:
Thanks @sleeptightAnsiC for your review. 
Sorry for the delay in the reply; it's a bit too hot in my house to program these days.
I will dedicate time to this project as soon as the temperature is more manageable here.
Also, thanks again for your https://github.com/skeeto/w64devkit/issues/251#issuecomment-4364942668, it has been helpful.

src/make2compdb.c
g-berthiaume: > The "gymnastic" around returning cwd is a bit odd. dirstack_pop is only used in this one place. 

I agree with you. I used to use this value to validate that we pop the expected directory, but for fuzzing, this made no sense.

Should be fixed in my next commit.

src/make2compdb.c
g-berthiaume: This is fixed in my next commit.
Thanks for your attention to detail.

src/make2compdb.c
g-berthiaume: Thanks for the suggestion.
I just tried it. To be honest, it feels to me like it would be a bit confusing for new users.
That said, I'm not opposed to this change.

I will commit it so you guys can try it, and I'll revert it if need be.

src/make2compdb.c
g-berthiaume: I like this discussion a lot.
I'm hesitant to fully commit to this idea because the core of my parsing strategy relies on finding the "start" of the compilation command, and therefore identifying the compiler.

Without it, I'm not sure how I would be able to navigate the noise that is a makefile stdout.

g-berthiaume:
## Round 3!

The focus for this new release is addressing the issues identified by @Peter0x44 and @sleeptightAnsiC.
Thanks for your help.

**Changelog**
- The `--verbose` mode is now piped to stderr.
- Simplify `CompilerKind` and add some new *"gcc-like"* compilers.
- Various fixes


**Note for reviewer**

As always, thanks again for your last feedback.
Please let me know if you see any areas where this tool could be improved.

**Previous rounds**
- Round 1: Initial proposal https://github.com/skeeto/w64devkit/pull/392#issue-4496079314
- Round 2: Tokenizer rework https://github.com/skeeto/w64devkit/pull/392#issuecomment-4627564473

g-berthiaume:
Hi @skeeto 
Just following up on this proposal. Are you still interested in this contribution?
If so, let me know if there’s anything missing or anything you’d like me to address.

skeeto:
Sorry, I did not prioritize a followup review enough and so I kept putting it off. It takes awhile to get my head back into the problem after being away from it more than a couple of weeks.

`str_unescape` applies C backslash semantics to shell lines. A backslash in POSIX shell means to take the next character literally, and so in a shell command `\t` isn't a tab but a literal `t`, with a superfluous backslash. For example:

    $ echo "gcc -c -Ds=\"It's fine\n\" x.c" | ./make2compdb.exe

Here the apostrophe is lost, and the `\n` becomes a newline where it should go into `gcc` as two bytes, backslash and `n`.

`clang` is missing a test, this `clang++` test fails, and the `nvc++` test uses the wrong string:

```patch
@@ -2558,2 +2558,5 @@ static void test_compiler_parse(Arena a)
 
+    run_test_compiler_parser(SL("clang"), SL("clang"), COMPILER_IS_GCC_COMPATIBLE);
+    run_test_compiler_parser(SL("clang++"), SL("clang++"), COMPILER_IS_GCC_COMPATIBLE);
+
     // zig cc
@@ -2573,3 +2576,3 @@ static void test_compiler_parse(Arena a)
     // NVIDIA HPC C++ compiler (nvc++ -std=c++17 -O2 -c main.cpp -o main.o)
-    run_test_compiler_parser(SL("nvc"), SL("nvc"), COMPILER_IS_GCC_COMPATIBLE);
+    run_test_compiler_parser(SL("nvc++"), SL("nvc++"), COMPILER_IS_GCC_COMPATIBLE);
 
```

`print_str_escaped_string`: The writer shouldn't produce `\v` nor `\a` (not valid JSON escapes), and all control characters should be escaped with at least `\u00xx`.

Should `str_drop_tail` clamp to `len - 1`? That doesn't make sense to me. Though it doesn't seem to cause any bugs.

Some typos in the header doc:

```patch
@@ -1,3 +1,3 @@
 // make2compdb
-// Generates a Clang's JSON Compilation Database files (`compiler_commands.json`) from make build systems.
+// Generates a Clang's JSON Compilation Database files (`compile_commands.json`) from make build systems.
 // This json file can be usefull for multiple tools including the [clangd](https://clangd.llvm.org/) LSP.
@@ -6,4 +6,4 @@
 //
-//      $ make -Bwn | make2compdb.exe > compiler_commands.json
-//      $ cat compiler_commands.json
+//      $ make -Bwn | make2compdb.exe > compile_commands.json
+//      $ cat compile_commands.json
 //      [
```

I think these are meant to go to standard error?

```patch
--- a/src/make2compdb.c
+++ b/src/make2compdb.c
@@ -2101,5 +2101,5 @@ static int make2compdb(Arena *perm, OsWriterInterface *os_stdout, OsWriterInterf
         else {
-            print_str(os_stdout, SL("Unknown CLI arg: '"));
-            print_str(os_stdout, arg->str);
-            println_str(os_stdout, SL("'\n"));
+            print_str(os_stderr, SL("Unknown CLI arg: '"));
+            print_str(os_stderr, arg->str);
+            println_str(os_stderr, SL("'\n"));
             return -1;
```

Also please exit with 1, not -1. That's a bad habit from the unix world where conventionally it's clamped to 255, but Windows uses the whole 32-bit exit status, and so it clamps to `0xffffffff`:

    $ echo 'int main() { return -1; }' | cc -xc -O - && ./a.exe
    sh: ./a.exe: Error 0xffffffff

The busybox-w32 shell reports such high-valued errors loudly because they're often significant (special Windows errors).

`make2compdb()` has a `static` counter variable, `s_prev_command_count`.

`print_compiler_invocation` appears to be missing a newline in its ouput.

Bad text replacement here?

```c
        goto reject; // unU8_paired low surrogate
```

* * *

I think we're pretty close to merging. `str_unescape` is an especially important blocker and needs to be fixed. It's also the most complicated. Everything that follows is in roughly decreasing priority.

g-berthiaume:
> @skeeto Sorry, I did not prioritize a followup review enough and so I kept putting it off. It takes awhile to get my head back into the problem after being away from it more than a couple of weeks.

No worries.
Since the last time we spoke was in May, I wanted to make sure my contributions were still aligned with your vision of the tool.

Thank you for your review; I'll address those problems shortly.

Dockerfile
```diff
@@ -962,6 +962,10 @@ RUN printf "id ICON \"$PREFIX/src/w64devkit.ico\"" >w64devkit.rc \
  && $ARCH-gcc -DEXE=../share/nsis/bin/makensis.exe -DCMD=makensis \
         -Oz -fno-asynchronous-unwind-tables -Wl,--gc-sections -s -nostdlib \
         -o $PREFIX/bin/makensis.exe $PREFIX/src/alias.c -lkernel32 \
+ && $ARCH-gcc \
+        -Oz -fno-asynchronous-unwind-tables -fno-builtin -Wl,--gc-sections \
+        -s -nostdlib -o $PREFIX/bin/make2compdb.exe $PREFIX/src/make2compdb.c \
```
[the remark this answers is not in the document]
g-berthiaume: @Peter0x44, your help would be appreciated here.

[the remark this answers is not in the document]
Peter0x44: I'm pretty sure the AI is just wrong here. The whole source directory gets copied, which would include make2compdb.c

src/make2compdb.c
```diff
@@ -0,0 +1,3527 @@
...
+        println_strlist(w, *line_tokens);
+        w->tab -= 1;
+    }
+
+    while (!strlist_is_empty(*line_tokens)) {
+        Str token = strlist_pop_front(line_tokens);
+
+        switch (state) {
+        case SEARCH_COMPILER: {
+            compiler = compiler_parse(token);
```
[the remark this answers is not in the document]
g-berthiaume: Solving this issue conflicts with our ability to handle arbitrary compiler wrappers.
I think I prefer generating a "fake command" (which is not a problem in all use cases for `compile_commands.json` I know of) if we see `echo gcc main.c` then dropping support for compiler wrappers.

I'll think about it.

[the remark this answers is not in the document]
Peter0x44: I wouldn't take this one too seriously. Do what you like, but it's no blocker.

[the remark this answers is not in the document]
g-berthiaume: Thank you Peter.

Dockerfile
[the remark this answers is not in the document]
g-berthiaume: Thanks !

src/make2compdb.c
```diff
@@ -0,0 +1,3527 @@
...
+
+    Str_Pair segment  = {.tail = input};
+    Compiler compiler = {.kind = COMPILER_IS_UNKNOWN};
+    while (segment.tail.len > 0) {
+        segment = str_cut(segment.tail, SL("-"));
+
+        if (str_equal(segment.head, SL("clang"))) {
+            compiler.kind   = COMPILER_IS_GCC_COMPATIBLE;
+            compiler.string = input;
+            return compiler;
```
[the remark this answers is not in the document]
Peter0x44: Honestly, idk if that's the right shape considering techncially gcc can have arbitrary names (--program-prefix or --program-suffix to configure). 
But maybe it shows up in Makefiles sometimes.
I don't think it's too big of a concern anyway.

g-berthiaume:
### Round 4

Here we go for round 4!
This round focuses on string escaping/unescaping and robustness.

@skeeto Every one of your comments should be addressed.
@Peter0x44 The automated review comments should also be addressed (I would be curious to see what a re-run would produce).

It took me longer than I'm willing to admit to wrap my head around c-string-escaping, JSON-escaping, and POSIX shell unescaping, and it looks like I'm not alone in finding this difficult: bear and compiledb, the two most popular implementations of make2compdb, both output incorrect JSON for this input:

```
gcc -c -Ds="It's fine\n" x.c
```

Here's the output of `compiledb`:


```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "gcc",
            "-c",
            "-Ds=It's finen", //< Look here
            "x.c"
        ]
    }
]
```

Here's the output of `bear`:


```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "gcc",
            "-c",
            "-Ds=It's finen",  //< Look here
            "x.c"
        ]
    }
]
```

Finally, here's the output of `make2compdb`:

```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "gcc",
            "-c",
            "-Ds=It's fine\\n",  //< Look here
            "x.c"
        ]
    }
]
```

This is a work in progress, but these are pretty encouraging results.

Speaking of problems: here's a fun one I encountered.
Did you know that posix-shell-unescaping is incompatible with Windows paths? 
Indeed. They are.

To illustrate this problem, here's an example from compiledb's [issue tracker](https://github.com/nickdiego/compiledb/issues/124):

```
g++ -IC:\VulkanSDK\1.3.216.0/include -c shader.cpp -o shader.o

```

Let's see how the tools handle this problem.

Here's the output of `compiledb`:

```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "shader.cpp",
        "arguments": [
            "g++",
            "-IC:VulkanSDK1.3.216.0/include",  //< Look here
            "-c",
            "shader.cpp",
            "-o",
            "shader.o"
        ]
    }
]
```

Here's the output of `bear`:

```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "shader.cpp",
        "output": "shader.o",
        "arguments": [
            "g++",
            "-IC:VulkanSDK1.3.216.0/include",  //< Look here
            "-c",
            "shader.cpp",
            "-o",
            "shader.o"
        ]
    }
]
```

What about `make2compdb.exe` ?

```json
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "shader.cpp",
        "output": "shader.o",
        "arguments": [
            "g++",
            "-IC:/VulkanSDK/1.3.216.0/include",  //< Look here
            "-c",
            "shader.cpp",
            "-o",
            "shader.o"
        ]
    }
]
```
🥳


Here's a non-exhaustive changelog:

- Rework JSON string escaping 
- Rework POSIX shell word unescaping
- Rework how we do testing: Add `run_test_shell_parse_line()` for higher-level tests
- Improve how we parse gcc's `-x`, aka the language flag
- Improve compiler detection
    e.g., does not detect `clang-format` and `gcc-ar` as a compiler
- Improve error handling for "Unknown CLI arguments"
- Improve OS stream writing
- Fix typos

### Note for reviewer

As always, thanks again for your last feedback.
I'm looking forward to your review.

#####  Previous rounds

- Round 1: Initial proposal https://github.com/skeeto/w64devkit/pull/392#issue-4496079314
- Round 2: Tokenizer rework https://github.com/skeeto/w64devkit/pull/392#issuecomment-4627564473
- Round 3: Compiler detection and cleanup https://github.com/skeeto/w64devkit/pull/392#issuecomment-5120391355

skeeto:
Great work, your changes completely address everything I mentioned. Again,
we're really close, but the changes brought some new, small things in.
Starting with the most important.

In-place normalization of `arg` interferes with the shared view in
`token`:

    $ echo 'gcc -IC:\\Users\\foo -c x.c' | make2compdb | grep foo
                "-IC:/Users/foooo",

Source tokens are escaped twice. Here `"file"` is correct but the last
argument is incorrect:

    $ echo "gcc -c 'It'\''s.c'" | make2compdb
    [
        {
            // ...
            "file": "It's.c",
            "arguments": [
                "gcc",
                "-c",
                "Its.c"
            ]
        }
    ]

The output file is not escaped. Here is last argument is correct but
`"output"` is incorrect:

    $ echo "gcc -c x.c -o 'my out.o'" | make2compdb
    [
        {
            // ...
            "file": "x.c",
            "output": "'my out.o'",
            "arguments": [
                // ..
                "-o",
                "my out.o"
            ]
        }
    ]

In all three cases you could fix it simply by unescaping once up front
into a copy.

After `os_stream_flush(win_bufstream_stdout);` the error flag should be
checked and exit status updated accordingly.

Some non-static functions slipped in:

    $ cc -c src/make2compdb.c
    $ nm make2compdb.o | grep '\<T\>'
    0000000000006286 T gcc_identify_language_from_x_flag
    // ...
    00000000000017fe T strbuf_add
    0000000000001766 T strbuf_alloc
    000000000000184c T strbuf_finalize

    $ cc -DTEST -c src/make2compdb.c
    $ nm make2compdb.o | grep '\<T\>'
    // ..
    0000000000008b64 T run_test_unescape
    // ..
    0000000000010e9d T test_shell_parsing

* * *

(The rest are things not blocking a merge.)

Matters less for w64devkit since we don't have aliases for these but some
Linux distribution cross toolchains put `-posix` and `-win32` suffixes on
tools:

    $ echo x86_64-w64-mingw32-gcc-posix -c x.c | make2compdb
    [
    ]
    $ echo x86_64-w64-mingw32-gcc-win32 -c x.c | make2compdb
    [
    ]

`cpp-output` should be classified as C, not C++. That's C Pre-Processor
"cpp". An unfortunately historical accident.

Very minor since it won't happen in practice, but a trailing `-o` invents
an argument:

    $ echo gcc x.c -o | make2compdb | jq .[0].arguments
    [
      "gcc",
      "x.c",
      "-o",
      ""
    ]

There's a duplicate `flag.kind == GCC_FLAG_IS_NOT_A_FLAG` branch.

Quick typo fixes:

    sed -i s/pck-config/pkg-config/ src/make2compdb.c

* * *

I expect you can fix all these quickly, and then I'd be satisfied to merge
your new tool.

g-berthiaume:
> @skeeto  Great work, your changes completely address everything I mentioned. Again, we're really close, but the changes brought some new, small things in.

No problem.
I contribute to this project because I like its high quality standard.
Also, I personally think of this project as a challenge, so your thorough reviews are welcome.

## Round 5

This round is marked by one change: we now escape all tokens.
Previously, I was hesitant at doing so for performance reason and my concern on handling windows path correctly.

Now that it's done, I'm pretty happy about the results.
In my benchmark, we remain the fastest tool, and the code is now easier to read and therefore maintain.

For the curious, here's a naive benchmark for my own project:
```txt
make2compdb      0.015 s
bear             0.020 s
compiledb-go     0.040 s
compiledb        0.465 s
```

Here are some of the fixes that you'll be able to see:

```sh
λ echo 'gcc -IC:\\Users\\foo -c x.c' | ./make2compdb
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "gcc",
            "-IC:/Users/foo",
            "-c",
            "x.c"
        ]
    }
]
```

```sh
λ echo "gcc -c 'It'\''s.c'" | ./make2compdb 
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "It's.c",
        "arguments": [
            "gcc",
            "-c",
            "It's.c"
        ]
    }
]
```

```sh
λ echo "gcc -c x.c -o 'my out.o'" | ./make2compdb
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "output": "my out.o",
        "arguments": [
            "gcc",
            "-c",
            "x.c",
            "-o",
            "my out.o"
        ]
    }
]
```

```sh
λ echo "x86_64-w64-mingw32-gcc-posix -c x.c" | ./make2compdb
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "x86_64-w64-mingw32-gcc-posix",
            "-c",
            "x.c"
        ]
    }
]
```

```sh
λ echo 'gcc x.c -o' | ./make2compdb
[
    {
        "directory": "C:/dev/gb/gab_w64devkit",
        "file": "x.c",
        "arguments": [
            "gcc",
            "x.c",
            "-o"
        ]
    }
]
```


Also, here's `peports` 

```txt
peports.exe make2compdb.exe 
KERNEL32.dll
        0       ExitProcess
        0       GetCommandLineW
        0       GetConsoleMode
        0       GetCurrentDirectoryW
        0       GetStdHandle
        0       ReadFile
        0       VirtualAlloc
        0       WriteConsoleW
        0       WriteFile
SHELL32.dll
        0       CommandLineToArgvW
```

And the `static` check:

```sh
gcc -c ./w64devkit/src/make2compdb.c
nm make2compdb.o | rg " T"
0000000000009b4f T mainCRTStartup

gcc -DTEST -c ./w64devkit/src/make2compdb.c
nm make2compdb.o | rg " T"
00000000000153bb T main
```

### Note for reviewer

I think we are approaching something pretty good!

#### Previous rounds:


- Round 1: Initial proposal https://github.com/skeeto/w64devkit/pull/392#issue-4496079314
- Round 2: Tokenizer rework https://github.com/skeeto/w64devkit/pull/392#issuecomment-4627564473
- Round 3: Compiler detection and cleanup https://github.com/skeeto/w64devkit/pull/392#issuecomment-5120391355
- Round 4: String escaping/unescaping https://github.com/skeeto/w64devkit/pull/392#issuecomment-5621078048

skeeto:
Thanks for the quick followup! Just a small regression on the `eol` change, which was a dead branch before:

    $ echo 'gcc -c x.c 2>&1'  | ./make2compdb.exe | jq -c .[0].arguments
    ["gcc","-c","x.c"]
    $ echo 'gcc -c x.c  2>&1' | ./make2compdb.exe | jq -c .[0].arguments
    ["gcc","-c","x.c","2"]

That branch just needs deleting, so I pushed f27e397 into your branch. I also pushed 0b6d72b since I'm ready to merge (with a big squash). Are you happy with me merging at 0b6d72b? I'm glad you did this, @g-berthiaume.

g-berthiaume:
Yes! Let's do it. I'm really happy to see this tool land. 🥳

I'll stick around for maintenance, so please feel free to tag me on any issues related to make2compdb.
And thanks again to @Peter0x44, @sleeptightAnsiC, and @skeeto for all the reviews and comments.

Gabriel

skeeto:
Forgot to push here first before I pushed to `master`, and it seems doing it the other way doesn't count as merge. Oh well. It was merged as 16ed33f. Thanks for doing this, @g-berthiaume!

g-berthiaume:
Happy to contribute! 
I have less time this fall, but I might have more time in the winter to do something else.
Thanks again for your mentorship, Chris. As I typically work in embedded, this was my first project with Win32, buffered streams, Unicode conversion, and string escaping. I've learned a lot!

Gabriel

Issue #326, opened 2026-02-13
`make`: mismatch in behavior launching `cmd`


WardBrian:
I am trying to build a project that uses an older version of the Threaded Building Blocks library that[ uses `cmd //C`](https://github.com/stan-dev/math/blob/98c8dcf32bf31677b3516075b59613c8d1a7f44a/lib/tbb_2020.3/build/windows.inc#L113-L114) in a few places in their makefile. 

This fails when using w64devkit, it appears to launch `cmd` and hang waiting for user input rather than running the command that follows. This behavior does not occur when using something like `RTools` or `msys` that provide similar environments for Windows.

It seems that `cmd /C` does work, but I believe that would break these other longstanding environments

---

As an aside, if I hack around these issues in TBB and try to proceed with the rest of my build, I get linker errors for standard library functions:

```
 undefined reference to `std::istream::seekg(std::fpos<_Mbstatet>)'
```

WardBrian:
I guess it might be useful to state my overall goal: I'd love to be able to download `cmdstan-2.38.0.tar.gz` from https://github.com/stan-dev/cmdstan/releases/tag/v2.38.0, untar it, and sucessfully run `make build`.

skeeto:
That's interesting, thanks for pointing this out. This is already handled in busybox-w32 as a special case for tools that assume an MSYS environment, mainly Autotools. However it's only for lowercase `//c`, because that's how Autotools spells it:

https://github.com/rmyorston/busybox-w32/blob/de7b34e8/shell/ash.c#L6954

It seems this ought to match `//C` as well. This is the first time I've seen a script use the uppercase version.

skeeto:
I just realized that's not quite enough because it's only looking for `echo`, too. (Again, this is mainly an Autotools un-hack.) 

The commands in that makefile strongly presume an MSYS environment. There is no virtual filesystem handling in w64devkit that will translate `//nologo`, etc. to `/nologo`.

WardBrian:
Thanks for investigating. If supporting these kinds of older builds is out of scope I understand!

avih:
> Thanks for investigating. If supporting these kinds of older builds is out of scope I understand!

It's not about being out of scope.

This command is invalid. Try running in cmd.exe prompt:
```
cmd //C echo 123
```
and you will see that it starts a new cmd.exe prompt, doesn't do the echo, and doesn't exit it (i.e. now you're in a nested cmd.exe shell).

[...]

PR #396, opened 2026-05-28
Mcfgthread

Peter0x44:
This is a branch of w64devkit with mcfgthreads, for anyone who wants to try them out.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -373,7 +373,9 @@ RUN case "$ARCH" in \
         --includedir=include \
         --libdir=lib \
  && meson compile -C /x-mcfgthread/build \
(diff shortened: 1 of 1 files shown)


Peter0x44:
@wardbrian Since I went to the trouble of making this branch, do you care to try if it works for your needs?

WardBrian:
I will try to give this an attempt in the next week. My primary project I was interested in using this for was also what led to #326, so I will need to get a bit creative

WardBrian:
@Peter0x44 I can't compile my project with this branch, getting a linker error ``undefined reference to `std::istream::seekg(std::fpos<_Mbstatet>)'``

Issue #409, opened 2026-08-02
The Severity and Facility values are being zero out on next message


danil-kondr2016:
When I trying to compile MC file with setting severity and facility value in first message and omitting them in subsequent ones, results are differing between MSVC's MC and windmc. The facility and severity values doesn't being copied from previous message like MC does.

<details>
<summary>msg.mc file</summary>

```
SeverityNames=(Success=0x0 Informational=0x1 Warning=0x2 Error=0x3)
FacilityNames=(
    Installer=0x701:PIW_FACILITY_INSTALLER 
    Parser=0x702:PIW_FACILITY_PARSER
)
LanguageNames=(English=0x409:MSG00409)

MessageId=0x1
Severity=Error
Facility=Installer
SymbolicName=PIW_E_PROGRAM_INCOMPATIBLE
Language=English
The program %1 is incompatible with target operating system.
.

MessageId=+1
SymbolicName=PIW_E_FAILED_TO_INSTALL
Language=English
Failed to install program %1.
.

MessageId=+1
SymbolicName=PIW_E_INSTALLER_NOT_FOUND
Language=English
The installer of program %1 not found at %2.
.

MessageId=0x1
Severity=Error
Facility=Parser
SymbolicName=PIW_E_SYNTAX_ERROR
Language=English
Syntax error at line %1!d!.
.

MessageId=+1
SymbolicName=PIW_E_PROGRAMS_NOT_FOUND
Language=English
There are no programs inside the program list.
.
```
</details>

<details>
<summary>msg.h produced by windmc</summary>

```
/* Do not edit this file manually.
   This file is autogenerated by windmc.  */

//
//  The values are 32 bit layed out as follows:
//
//   3 3 2 2 2 2 2 2 2 2 2 2 1 1 1 1 1 1 1 1 1 1
//   1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0 9 8 7 6 5 4 3 2 1 0
//  +---+-+-+-----------------------+-------------------------------+
//  |Sev|C|R|     Facility          |               Code            |
//  +---+-+-+-----------------------+-------------------------------+
//
//  where
//
//      C    - is the Customer code flag
//
//      R    - is a reserved bit
//
//      Code - is the facility's status code
//
//      Sev  - is the severity code
//
//           Success - 00
//           Informational - 01
//           Warning - 02
//           Error - 03
//
//      Facility - is the facility code
//
//           System - 00ff
[...]

PR #410, opened 2026-08-02
binutils: patch windmc to save previous severity and facility

danil-kondr2016:
Patch for fixing of #409


skeeto:
Thank you for #409, but your fix here in #410 is a no-op and does not change the output. These global stores are copied into a struct but then never read. I've added be006bc to `windmc-bom` to fix #409. (CI is failing due to a timestamp issue I need to sort out.)

While testing against `mc.exe` Opus noticed a related facility counting difference, so  I also have 860b529. The same investigation also found a bug in `mc.exe` where it doesn't match the documented behavior (default Application value), but `windmc` does, so it's possible with these patches `windmc` is now the less buggy of the two message compilers!

PR #417, opened 2026-09-06
Re-enable LTO, create fat LTO C++ and Fortran runtimes

skeeto:
This branch is a serious exploration of re-enabling LTO and shipping **fat LTO** libstdc++, libgfortran, and libquadmath. My personal motivation is to produce better C++ binaries: smaller and faster. In this branch LTO is opt-in and so costs nothing (other than installation size) unless enabled with `-flto` at both compile and link time. C++ and Fortran binaries compiled with LTO enabled are significantly smaller, probably faster (TODO on benchmarks), and users aren't locked into `-O2`, though the link-time cost is substantial.

This branch includes patches that improve and fix a number of issues with LTO in GCC and Binutils. Cross-compiled LTO doesn't work out-of-the-box with GCC due to bytecode generation bugs in GCC (w64dk may be the first project to ever attempt it), patched in this branch by necessity.  In general, most of the effort of this branch has been playing whack-a-mole with GNU toolchain LTO bugs. That's exactly why I disabled LTO in the first place: It's difficult to trust that any given binary linked using GNU LTO is built correctly!

Concurrency runtimes are deliberately excluded from LTO: winpthreads, libgomp, and libatomic. I know for a fact that winpthreads is broken with LTO, and I will not trust the others with LTO without and audit (TODO).

The business with `make` is because `-flto=auto` integrates with it (jobserver).


Peter0x44:
Last I checked, gcc couldn't build itself at -O2 (for windows) with lto without giving an ICE. Hopefully that changed.

skeeto:
I didn't think building GCC itself with LTO would be worth much except as a way to detect LTO bugs earlier. Worth at least testing, though.

